PR #117, opened 2014-10-06
Clean up the code by using the Variation flag 'include_failed_variations...

ens-lgil:
...(1)' in the Hub module, instead of setting the flag in every module where a Variation adaptor is created


ens-ds23:
#143

PR #143, opened 2014-11-24
Fix to include flagged 'failed' variants in the variation tables

ens-lgil:


ens-ds23:
#117

ens-ds23:
It turns out this wasn't the right thing to do. Instead it was an issue with the incompleteness of #117 . Recently further bugs have been revealed relating to that incompleteness. I'll shortly make a commit which effectively reverts this pull and corrects #117 .

PR #157, opened 2015-01-14
Preventing code from failure in case variation is undef. Lots of EG spec...

jkhobova:
In EG some species have no variation DB. Changed to call include_failed_variation only in case variation db exists.


ens-hsr:
Do you need this fix on release/78 too?

jkhobova:
yes, we do.

Thank you,
Julia

ens-hsr:
Could you send this pull request against postreleasefix/78 branch please?

ens-hsr:
Manually merged into both master and release/78: 3757294bc4de52625ba542819bf23fc93f99cbb5

jkhobova:
thank you!

PR #175, opened 2015-02-20
get_all_ConsequenceType API method no longer exists and appears to be unused here

nicklangridge:


ens-lgil:
Hi Nick,

This is the same fix in Will's pull request #167 "remove reference to unused and now removed class/method".

Cheers,
Laurent

PR #183, opened 2015-02-27
Clean some parts of the code and remove a non useful loop

ens-lgil:


ens-ds23:
Hi Laurent, The loop you removed seems to sort the array by the order of any leading digits, whereas your code now relies on the order in the array. Are you saying that this is unnecessary or incorrect, or is it an oversight? Dan.

ens-lgil:
Hi Dan,

The HTML table already sorts the evidences by position or breakpoint order.
Anyway there is no need to sort the evidences by their ID and it 
slightly speeds up the page.

Cheers,

Laurent

ens-ds23:
Thanks, Laurent! Harpreet, it's ok to merge this in when you're ready tomorrow. @ens-hsr

PR #184, opened 2015-02-27
Change the columns order to match the new columns order of the Gene/Phen...

ens-lgil:
...otype page


ens-ds23:
Thanks for the code tidying as well.

PR #190, opened 2015-03-06
Add the population description: this will appear when the user move the ...

ens-lgil:
...mouse on the population name. Update the table display (i.e. no header) for the somatic data


ens-hsr:
Could you please set it against postreleasefix/79?

PR #195, opened 2015-03-12
When choosing nearest feature, scale the window according to genome size

nicklangridge:
Otherwise the window may be too big to display for small genomes


ens-ds23:
Wow, this is old: sorry about that. Looking into it now.

PR #202, opened 2015-03-31
Update the rat RGD link

ens-lgil:


ens-lgil:
Would it be possible to copy the new URL to the branch release/79 as well, please ?

ens-hsr:
Please send it against postreleasefix/79 in that case.

ens-lgil:
It seems that the DEFAULT.ini file has differences between the 'master' and 'release/79' branches.
It looks like it's ok to overwrite the differences in 'release/79'

PR #203, opened 2015-03-31
Update the rat RGD link

ens-lgil:
And merge a couple of new links from the 'master' branch


ens-st3:
Hi - this overwrites a couple of other changes - ZFIN_ID and DDG2P. I know the first should not be overwritten, do you know about the second ?

ens-lgil:
The other changes come from the version on release/79 and hasn't been committed to the master branch.
This changes happened when I merged the versions on master and release/79.

ens-st3:
Neither of those changes are in release/79 but are in the version on master (3f6f9b47), so it's a bit odd that the pull request is wanting to merge them again. They should in fact be overwritten however so no harm done.

PR #222, opened 2015-04-30
Add an ID for the sub-population blocks

ens-lgil:


modules/EnsEMBL/Web/Component/Variation/PopulationGraphs.pm
```diff
@@ -162,13 +164,13 @@ sub content {
...
     $html .= sprintf('
-      <div class="population_freq_%s population_genetics_pie">
+      <div class="population_freq_%s population_genetics_pie" id="population_freq_%s">
         <div class="toggleable" %s>
           <div><p><b>%s sub-populations</b></p></div>
           %s
         </div>
       </div>
-    ', $sp, $show ? '' : 'style="display:none"', $sp, $sub_html);
+    ', $sp, $sp, $show ? '' : 'style="display:none"', $sp, $sub_html);
```
ens-ma7: Hi Laurent
Looks like you have an extra $sp in here, is that ok? or is the diff which is confusing.

Ridwan

ens-lgil: Hi Ridwan,

This is because I added an extra '%s' (i.e. 'id="population_freq_%s') in the line 167.

ens-ma7:
Ah ok makes sense it was hard to spot this in the diff.

PR #223, opened 2015-04-30
Update the columns and the content of the columns so the render of the t...

ens-lgil:
...ables will be neater in the Ensembl mobile website

### DIFF
--- a/htdocs/components/85_spreadsheet_table.css
+++ b/htdocs/components/85_spreadsheet_table.css
@@ -122,3 +122,10 @@ td:hover a.hover_show  { display: inline; }
 
 tr.row-sub td { font-size: 85%; }
 tr.row-sub td:first-child { text-indent:24px; }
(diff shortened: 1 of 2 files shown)


ens-hsr:
There seems to be code repetition for Alleles and Genotypes. Can this be split out as a subroutine? Also please avoid using inline styles. Add any styles to an appropriate .css file in htdocs/components.

ens-ap5:
What Harpreet said. There’s documentation on where to find the pertinent styles here:

https://www.ebi.ac.uk/seqdb/confluence/display/ENS/Ensembl+web+development <https://www.ebi.ac.uk/seqdb/confluence/display/ENS/Ensembl+web+development>

Cheers

Anne

ens-lgil:
I moved the inline styles into the htdocs/components/85_spreadsheet_table.css file.

PR #235, opened 2015-05-29
Allow hubs to be attached using UCSC_GOLDEN_PATH and ASSEMBLY_VERSION

brucebolt:
Under the current code, a TrackHub can be attached using the value of UCSC_GOLDEN_PATH or ASSEMBLY_VERSION.  However, if UCSC_GOLDEN_PATH is defined for the species, the image config will not show any tracks from hubs using the ASSEMBLY_VERSION (even though the hub shows as correctly attached).  This change allows configuration of tracks where the ASSEMBLY_VERSION is being used in the hub.

This was causing a problem for C. elegans users where ce10 and WBcel235 are both commonly accepted assembly names (even though UCSC's ce10 assembly is considerably older than WormBase's WBcel235...), so users may be using either in their TrackHubs, but most likely to be using the (correct) WBcel235 value.

### DIFF
--- a/modules/EnsEMBL/Web/ImageConfig.pm
+++ b/modules/EnsEMBL/Web/ImageConfig.pm
@@ -813,9 +813,10 @@ sub _add_datahub {
 
     my $menu     = $existing_menu || $self->tree->append_child($self->create_submenu($menu_name, $menu_name, { external => 1, datahub_menu => 1 }));
 
+    my $assembly_default = $self->hub->species_defs->get_config($self->species, 'ASSEMBLY_VERSION');
     my $assembly = $self->hub->species_defs->get_config($self->species, 'UCSC_GOLDEN_PATH')
-                        || $self->hub->species_defs->get_config($self->species, 'ASSEMBLY_VERSION');
-    my $node = $hub_info->{'genomes'}{$assembly}{'tree'};
+                        || $assembly_default;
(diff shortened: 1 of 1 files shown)


ens-ds23:
Hi Bruce,

We've been meaning for some time to support aliases for assemblies on hubs, and I should get on with it as it's hitting people. I'll add a configuration array to allow an arbitrary list of acceptable aliases in trackhubs.

By the way, do you have or know of a good list of correspondences between assembly names for worm?

brucebolt:
Hi Dan

WBcel235 is the most commonly used assembly name.  This assembly has no corresponding UCSC name, as it is not hosted on their genome browser.

ce10 is the latest version available on UCSC, but this corresponds to an older assembly (WBcel220 IIRC), so is not a suitable alias.

Additionally, WormBase give each release an identifer.  The current version in Ensembl is WS245, which uses the WBcel235 assembly.

ens-ds23:
Ah that's great, thanks. I'm just writing the patch now. So, as I understand it, we should probably support aliases called WS245 and WBcel235 in hubs, and no UCSC name, as worm is currently released?

brucebolt:
WS245 and WBcel235 should be fine as the aliases.

ens-ds23:
Fixed by allowing arbitrary aliases at ENSWEB-1284.

ens-ds23:
(Thanks for the patch, though. That definitely speeded up implementing the final fix).

brucebolt:
Thanks for implementing this Dan.  Should solve the problems we've been having.

PR #240, opened 2015-06-22
add new var icon and code to use it

willmclaren:


ens-hsr:
Will, can you please send the pull request again. It has got an extra branch-merge commit.

PR #241, opened 2015-06-22
add new var icon and code to use it

willmclaren:


ens-hsr:
Is 96/var_individual_genotypes.png still be used somewhere, or shall I remove that?

willmclaren:
There's still one page from the docs plugin using it, but that should be updated too.

You can remove it.

PR #245, opened 2015-06-25
At7/feature 81/ld bug

at7:
Use the correct array for counting the number of variants.


ens-hsr:
Could you please remove the two extra commits?

at7:
Done.

PR #250, opened 2015-07-01
Update TextSequence.pm

fcunningham:
Changed 'variations' to 'variants' in messages sections. Also wonder if we can do this on line 1297 but didn't want to without creating a sandbox.


fcunningham:
Please wait for another 4 commits to come in from me.

fcunningham:
These are all in now (even if there aren't 4)

PR #259, opened 2015-07-13
Changed variation to variant

fcunningham:
I'll stop now for a while! Hope these are ok.


ens-st3:
Can you submit a pull request against release/81 (assuming we want it in this next release) ?

PR #263, opened 2015-07-21
Create hyperlink for the variants listed in the LD table and highligh…

ens-lgil:
…t the focus variant


ens-lgil:
This change has been asked by the outreach team.

PR #284, opened 2015-09-02
genome_db.assembly_default has been removed in e82

muffato:
Also, we can simply take all the GenomeDBs regardless of their first/last_release status

### DIFF
--- a/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -1178,7 +1178,7 @@ sub _summarise_compara_db {
 
   ###################################################################
   ## Section for storing the genome_db_ids <=> species_name
-  $res_aref = $dbh->selectall_arrayref('SELECT genome_db_id, name, assembly FROM genome_db WHERE assembly_default = 1');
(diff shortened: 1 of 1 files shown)


muffato:
BTW, this function only does direct SQL queries on the database handle. Is it possible to have a Compara DBAdaptor instead ? We could then move the queries to our side and it would be easier to maintain

ens-ap5:
Hi Matthieu

If you want to have them in the API, that would be fine. We do use a lot of SQL in ConfigPacker because it’s grabbing things like track summaries that are not needed anywhere else and hence have historically not been incorporated into the respective APIs, but it does make the code harder to maintain.

Cheers

Anne

muffato:
ah, you think it would make things harder :)
My question mainly is whether it is possible to have a DBAdaptor. We don't need to move the code, but making API calls would make the ConfigPacker more resilient to schema changes (and slightly slower)

ens-ap5:
I meant that using SQL makes ConfigPacker harder for the webteam to maintain, because we have to keep up with schema changes.

Anne

muffato:
right, we're on the same line, then :)
This time, I just forgot to check if the schema was directly used by other repos, but I'll add a note to our SOP so that we don't miss it next time

PR #307, opened 2015-09-28
The Variation 'Genomic context' and 'Flanking sequence' pages are ava…

ens-lgil:
…ilable if the variant has at least one mapping


ens-hsr:
Is it for ENSWEB-1739?

ens-lgil:
Yes

Laurent

PR #311, opened 2015-09-30
remove redundant variation table configuration

willmclaren:


ens-ds23:
Hi Will,

It looks like I might be able to restore the intron context configuration now (that's why I was holding off on this). I'll merge this in and then restore just that option and push it live with the other outstanding fixes.

ens-ds23:
Accepted in 373c573 , but due to git-ness it was easier just to apply it as a patch there than merge.

PR #320, opened 2015-10-21
New Linkage Disequilibrium (LD) data track, focused on a variant inst…

ens-lgil:
…ead of a slice.
Use a new 'Manhattan plot' display for this track, using the new separate way to fetch the data and draw the track (i.e. the 'Style' repository).


ens-lgil:
The rest of the modules (i.e. the configuration and the content modules) will arrive later.
Anne, feel free to update/refactor the code.

ens-ap5:
OK, thanks! I'll look at it when I come back from holiday - much to do before then!

PR #349, opened 2015-12-04
Use the method 'markup_variation' from the module TextSequence instea…

ens-lgil:
…d of FlankingSequence


ens-lgil:
Thanks Steve!
I forgot to mention that I changed this because the highlight of the variants wasn't working in Flanking sequence, due to this old "markup_variation" method.

PR #351, opened 2015-12-04
Fix display issue by replacing the markup category 'variations' by 'v…

ens-lgil:
…ariants'


ens-ds23:
Confirmed. Thanks for fixing this.

PR #369, opened 2016-02-03
ENSEMBL-4232 - 'BLAST this sequence' strips out the FASTA ID

sanjay-eg:


sanjay-eg:
This pull request is in combination with the pull request https://github.com/Ensembl/public-plugins/pull/47

ens-hsr:
Hi Sanjay, could you try using '_seq' class instead of 'text_sequence_identifier'? You probably won't need https://github.com/Ensembl/public-plugins/pull/47 then.

sanjay-eg:
Yup.. done. Thanks.

PR #373, opened 2016-02-08
Removed unused/obsolete code as it is causing error on EG

sanjay-eg:
Could you merge it to Master and release/83 please?
That variable '$current_taxon' is not being used anywhere else in this module and is also causing issues on live EG sites.

Thanks,
Sanj


sanjay-eg:
Thanks @ens-hsr

PR #376, opened 2016-02-12
Add coverage glyphs in Variation and update it in StructuralVariation

ens-lgil:
The feature coverage glyph has been updated in the "Genes and regulation" page for StructuralVariation and has been added for the "Genes and regulation" page as well in Variation.


ens-st3:
Hi Laurent,

Can you send me a link to a sandbox so that I can see what this looks like ?

PR #378, opened 2016-02-15
Export this image bug - Broken on live

sanjay-eg:
This is broken on live on both E and EG. Please see the following page:
http://www.ensembl.org/Mus_musculus/Transcript/Population/Image?db=core;g=ENSMUSG00000041143;r=4:138972905-139059167;t=ENSMUST00000050949

This update seems to fix the issue on EG. Not sure how to test this fix against Ensembl site.

Thanks,
Sanjay


ens-st3:
@ens-lgil OK with you ?

ens-lgil:
Hi Steve,
I am ok with the fix

ens-st3:
thanks sanjay

PR #381, opened 2016-02-19
Consider improving visibility of rna tracks

nicklangridge:
For EG and VectorBase it was requested to improve the visibility of rna tracks as the contrast was not good on some screens. We changed the colour from hotpink2 to royalblue4.

I thought Ensembl might like to do the same if it doesn't cause any colour conflicts.

See screenshots.

Old (hotpink2)
old-hotpink2

New (royalblue4)
new-royalblue4


ens-hsr:
Merge at bf0fea31d1342224774d57896467c651bf2c7424

PR #399, opened 2016-03-07
Fixing BED. Missed conversion to 0-based start

andrewyatz:
Seems we did not minus 1 from our output when we were exporting into BED. Since BED is 0-based start and 1-based end we only need to minus this from the start. We assume all input coordinates are 1-based (they're from the API so this should be right).

### DIFF
--- a/modules/EnsEMBL/Web/Object/Export.pm
+++ b/modules/EnsEMBL/Web/Object/Export.pm
@@ -805,6 +805,8 @@ sub feature {
   }   
   if($format eq 'bed'){
     @mapping_result = qw(seqid start end name score strand);
+    # move coords into zero-based start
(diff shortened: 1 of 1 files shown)


andrewyatz:
Is it too late to get this onto release/84?

ens-ap5:
No, we can cherry-pick it onto the branch or whatever our paladin thinks best :)

Anne

andrewyatz:
:+1:

andrewyatz:
I'll check once this is on test then to make sure http://www.ensembl.org/Rattus_norvegicus/Export/Output/Location?db=core;flank3_display=0;flank5_display=0;output=bed;r=X:612525-712525;strand=feature;param=variation;_format=Text renders correctly

PR #404, opened 2016-03-10
Fix a couple of minor display issues for the Variation and StructuralVariation 'Genes and regulation' pages

ens-lgil:


ens-hsr:
Apologies Laurent, Ridwan has pushed something to master that should not go on release, so this pull request needs to be on postreleasefix/84 again please.

PR #407, opened 2016-03-14
set tmpdir to web tmpdir when using remote VCFs

willmclaren:
This is a change to support e!Genomes using the VCF backend with FTP-hosted VCFs.

The relevant support for this is only on master in ensembl-variation (not release/84), so only committing to master here.

@nicklangridge do you need this for release/84 too???


nicklangridge:
Yes please - if it can also go to release/84 that would be appreciated.

willmclaren:
I can either put in another PR, or whoever merges this can do whatever the thing is they need to do with post-release fixes.

I'll put the variation part on release/84

ens-hsr:
I have done "the thing" 20d037c4

PR #414, opened 2016-03-18
Use the hit description before looking for interpro description

nicklangridge:
Requested by @markmcdowall
Seems reasonable, but not sure if Ensembl has specific reason to avoid the hit description.


nicklangridge:
Any thoughts on merging this? (more context here if needed https://www.ebi.ac.uk/panda/jira/browse/ENSEMBL-4369)

ens-hsr:
Does it need to go live?

nicklangridge:
85 is fine

PR #417, opened 2016-04-15
Remove the 'moltype' method (dropped from the Variation SQL schema an…

ens-lgil:
…d API)


ens-ap5:
This method wasn't being used anyway - it hasn't been touched since we changed from SNP to Variation in our namespace way back in 2008!

PR #421, opened 2016-04-29
feat(Ontology): add "View associated genes" link if we don't have chr…

Zhicheng-Liu:
…omosome

Previously in Ensembl, where we have a chromosome, the ontology table offers a link to view xrefs on the karyotype, e.g.'View on karyotype'.

This change offers a link ("View associated genes") to the whole genome view if we don't have a chromosome.

This change also shows titles like "Locations of Genes associated with GO:0005634 nucleus" even if there is no karyotype image.

Related to ENSEMBL-4466.


ens-ma7:
As discussed with zhicheng and Nick, its better to have title "Gene information" above the table changed to "Genes associated with  .... " when there is no karyotype.

Closing this pull request and zhicheng will send a new PR.

PR #423, opened 2016-05-03
feat(Ontology): add "View associated genes" link if we don't have chr…

Zhicheng-Liu:
…omosome

Previously in Ensembl, where we have a chromosome, the ontology table offers a link to view xrefs on the karyotype, e.g.'View on karyotype'.

This change offers a link ("View associated genes") to the whole genome view if we don't have a chromosome.

This change also shows titles like "Locations of Genes associated with GO:0005634 nucleus" for the gene table when there is no karyotype image.

Related to ENSEMBL-4466.


ens-ma7:
Some of the code needs updating which zhicheng is looking into.

PR #425, opened 2016-05-05
Feature/table export

jyothishnt:
Main highlights:
  Highlight track on hover
  Track menu hover changed to click
  Userdata highlight
  Remove legend hover


ens-ap5:
Have we had the go-ahead from higher up to put this into release 85? There are quite a lot of changes here!

PR #429, opened 2016-05-18
ENSEMBL-4473 fix off-by-one VCF feature bug

nicklangridge:
Remove +1 as both positions are already 1-based


nicklangridge:
I'm fairly sure the +1 was added by mistake - according to wikipedia https://en.wikipedia.org/wiki/Variant_Call_Format the VCF POS field is 1-based, and the Ensembl slice start value should also be 1-based

ens-ap5:
I agree - it looks like that +1 has been there since the glyphset was first written, but the coordinate calculation has changed so it may have made sense originally.

PR #430, opened 2016-05-20
Fix intermittent 'no data' bug for population image/table

nicklangridge:
More details in JIRA ENSEMBL-4500

This fix configures the variation db to use VCF files if that is what is configured for the species - without it we will sometimes get adaptors that don't use the VCF and so return no data.

It's a pragmatic fix; it feels like a hack to put this configuration in the components and better solution might be to do it in a generic way at the time the adaptors are created. But after talking with @ens-ds23 it sounds like this approach has been tried before and is difficult to get right.


nicklangridge:
The fix is based on the similar code here https://github.com/nicklangridge/ensembl-webcode/blob/postreleasefix/84/modules/EnsEMBL/Web/Object/Transcript.pm#L1550

PR #445, opened 2016-06-30
fix(variation view): display the number of unique transcripts that a …

Zhicheng-Liu:
…variant overlaps

For multi allelic variants, the Gene and Transcript consequences table presents every overlapping transcripts for each allele. In the variation summary section we should show a variant overlaps X (unique) transcript(s).

Fix ENSEMBL-4480.


Zhicheng-Liu:
https://www.ebi.ac.uk/panda/jira/browse/ENSEMBL-4480

ens-st3:
thanks

PR #452, opened 2016-07-05
add ontology term tab 

sarahhunt:
This adds a table of mapping ontology terms to the locations page. It is written to be extended for e!86


ens-st3:
If it's not needed for 85 we'll merge it after we branch

sarahhunt:
We'd like to add tab this in e!85 then add further ontology support for e!86

PR #456, opened 2016-07-06
Update the variation and structural variation ZMenu tracks for the pr…

ens-lgil:
…ivate variation data (DECIPHER and LOVD)


ens-st3:
Thanks

PR #462, opened 2016-07-12
Avoid spamming logs with 'undefined' errors when static server is not…

nicklangridge:
… configured


ens-hsr:
7320f839402e7ff6f26220bbd95be95be8478b19

PR #464, opened 2016-07-12
Code from Will Spooner to improve sub population_structure.

dbarrell:
Instead of using hardcoded regex for ':ALL', now looks for leaf node of population hierarchy and works it's way up. Therefore will work for population structures of varying depth.

### DIFF
--- a/modules/EnsEMBL/Web/Component/Transcript/Haplotypes.pm
+++ b/modules/EnsEMBL/Web/Component/Transcript/Haplotypes.pm
@@ -179,19 +179,19 @@ sub short_population_name {
 sub population_structure {
   my $self = shift;
   my $pop_objs = shift;
-  
   if(!exists($self->{_population_structure})) {
     my %pop_struct;
     foreach my $pop(@$pop_objs) {
-      next if $pop->name =~ /:ALL$/;
-      my $subs = $pop->get_all_sub_Populations();
-      next unless $subs && scalar @$subs;
-      @{$pop_struct{$pop->name}} = map {$_->name} @$subs;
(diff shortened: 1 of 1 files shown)


ens-st3:
Hi Dan / @willmclaren, is this for release 85 ? Steve

willmclaren:
We don't need it, so I guess up to @dbarrell ?

dbarrell:
You are right - you only need it if you want to display sub populations from a species that does not have the exact same sub population structure as human does, so ensembl doesn't _need_ it whilst you only have transcript haplotypes for human. However this makes it generic to any sub population depth so it would be good to merge in for the future 🔮

ens-ds23:
Ok, I'll take this diff, merge it into master now (as it's on a very wrong branch because of the delay), and close the pr.

ens-ds23:
Ok, I've taken a look at the code and @willmclaren and @ens-ap5 have extensively modified the point at which the patch would be applied, so it's no longer a no-brainer. Does this still need applying? Is it worth reworking the patch?

dbarrell:
Hi Dan (S) Dang looks like I missed this. It would be really good if it could be applied else I'm going to have to re-write this at every release and make another PR

ens-ap5:
We can't merge this pull request directly because of Will's changes (mine are irrelevant, they're just caused by moving this module from its original plugin). Since I'm not familiar with the underlying data and API, I think @willmclaren  needs to look at it. It's only a few lines so could be done manually with little effort.

ens-ap5:
Will has written a compatible replacement.

dbarrell:
Thanks Folks!

PR #467, opened 2016-07-20
 made changes to the protein tree stats page to now include the Gene …

waakanni:
I have made changes to the protein tree stats page to now include the Gene QC stats


ens-ma7:
The changes have been pushed to the live server. Can you please check your changes on the live site to make sure nothing is broken. let me know if there is any problem.

PR #471, opened 2016-07-25
fix(Export): check variation db exists before accessing it

Zhicheng-Liu:
Fix ENSWEB-2791 and ENSEMBL-4593.


Zhicheng-Liu:
Closing this pull request as suggested by Harpreet this fix should go into postreleasefix/85.

PR #492, opened 2016-10-12
Group entries with the same phenotype in 1 row

ens-lgil:
Could be merge with the branch release/86 as well


ens-lgil:
Resubmitted to the branch postreleasefix/86

PR #496, opened 2016-10-19
Add 'sub_type' as a valid config item back to EnsEMBL::Draw::GlyphSet…

Zhicheng-Liu:
…::flat_file

Since this change (https://github.com/Ensembl/ensembl-webcode/commit/f2f2654b8520b67fb5bea028406316708f676094),
the 'sub_type' config item has been replaced with 'type'. But there are still places
where 'sub_type' is used instead.

Fixes ENSEMBL-4628.


Zhicheng-Liu:
If this change is fine, could you please also apply it to the release/86 and master branches?

ens-ds23:
I'll apply this as it looks fine: if that parameter doesn't exist, it should have no effect. But I don't really understand this area of code at all. Could I ask if @Zhicheng-Liu and @ens-ap5 could talk together when Anne is back to check that it really is ok, or at least so that Anne knows? Thanks!

PR #498, opened 2016-10-20
Gene-tree stats

muffato:
I addressed several comments from Paul K. and fixed http://www.ensembl.org/info/genome/compara/prot_tree_stats.html?page=coverage
All tested on my sandbox http://enssand-02.internal.sanger.ac.uk:9073/info/genome/compara/prot_tree_stats.html

### DIFF
--- a/modules/EnsEMBL/Web/Document/HTML/Compara/GeneTrees.pm
+++ b/modules/EnsEMBL/Web/Document/HTML/Compara/GeneTrees.pm
@@ -19,7 +19,7 @@ limitations under the License.
 
 package EnsEMBL::Web::Document::HTML::Compara::GeneTrees;
 
-## Provides content for compara gene-trees documentation
+## Provides content for compara gene trees documentation
 ## Base class - does not itself output content
 
 use strict;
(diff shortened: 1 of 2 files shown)


ens-ds23:
Checked all commits except 93a.

ens-ds23:
93a is fine, too.

ens-ds23:
Do these need to be pushed out for 86, @muffato ?

muffato:
Yes please ! Thank you @ens-ds23

muffato:
Hi Dan,
Has it been pushed live ? I still get the error on http://www.ensembl.org/info/genome/compara/prot_tree_stats.html?page=coverage

ens-ds23:
Hi @muffato , not yet. There was a bit of a backlog on Thursday with stuff needing to be pushed live, and then I had to unexpectedly take a day off yesterday. I'll push it today.

muffato:
No worries, just wanted to check if I had missed a bug

ens-ds23:
Hi Matthieu,

Working now. It looks very neat! We'll have to steal those little 
pie-charts for some of our other tables, :-) .

Dan.

muffato:
To be honest, these piecharts are made with what you'd probably call "old technology", i.e. Raphael. When I introduced these charts ~2 years ago, I just copied what had been chosen for the LastZ stats. But adding a lot of piecharts to a page is a PITA:
 1. You can only have 1 colour space for the whole page, so if you want some yellow+blue piecharts and some red+green ones on the same page you need to define a palette with the four colours and make piecharts with values such as [a,b,0,0] or [0,0,c,d]
 2. For each piechart you need to define an <input> with the data values and a <div>placeholder for display. The <div> have to be explicitly indexed with 0..n to be linked to the each <input> (the <input> are implicitly indexed by their order in the HTML document). So here we need to have a shared variable counting the number of charts already placed. https://github.com/muffato/ensembl-webcode/blob/master/modules/EnsEMBL/Web/Document/HTML/Compara/GeneTrees.pm#L163
I think Variation is using another library that may be easier to work. Check with them first :)

PR #499, opened 2016-10-21
Initialise an empty species list if not given

Zhicheng-Liu:
This was causing an JS error for EG Bacteria which prevents its genome search
box working. Although in Bacteria this panel may not be created, this change
provide a quick fix.


Zhicheng-Liu:
If this change is fine, could you please also apply it to the release/86 and master branches?

PR #505, opened 2016-11-14
To accommodate EG admin plugin's overwrite of ensembl-orm plugin

sanjay-eg:


sanjay-eg:
Pull req: https://github.com/Ensembl/ensembl-webcode/pull/503
merged into master. So wont need this pull request.

PR #512, opened 2016-11-25
Matthieu misc compara

muffato:
Two bugfixes:
 - ENSCOMPARASW-873. Matthieu Barba complained that the colour of our piecharts were not legible
 - ENSWEB-3397

All tested on my sandbox
http://enssand-02.internal.sanger.ac.uk:9073/info/genome/compara/analyses.html
vs
http://test.ensembl.org/info/genome/compara/analyses.html

and
http://enssand-02.internal.sanger.ac.uk:9073/info/genome/compara/prot_tree_stats.html
vs
http://test.ensembl.org/info/genome/compara/prot_tree_stats.html

Matthieu


muffato:
Thank you !

PR #517, opened 2016-12-06
Check datatype of validation status

nicklangridge:
Can't assume $vstatus is an array


nicklangridge:
Please can this also get cherry-picked to v86?

PR #523, opened 2016-12-16
ENSEMBL-4710 Fix pointy start of chromosome

MatBarba:


ens-ap5:
Done - please ask EG web to push your amended file out to their websites.

PR #524, opened 2017-01-04
Use Bio::EnsEMBL::Gene object to get colour key

Zhicheng-Liu:
Instead of EnsEMBL::Web::Object::Gene ...

Fixes ENSEMBL-4805.

### DIFF
--- a/modules/EnsEMBL/Draw/GlyphSet/TSE_transcript.pm
+++ b/modules/EnsEMBL/Draw/GlyphSet/TSE_transcript.pm
@@ -39,7 +39,7 @@ sub render_normal {
   my $coding_start      = $trans_obj->{'coding_start'};
   my $coding_end        = $trans_obj->{'coding_end'};
   my $transcript        = $trans_obj->{'transcript'};
-  my $colour_key        = $self->colour_key($config->core_object('gene'), $transcript); # need both gene and transcript to get the colour
+  my $colour_key        = $self->colour_key($config->core_object('gene')->Obj, $transcript); # need both gene and transcript to get the colour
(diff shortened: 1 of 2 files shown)


ens-st3:
Hi - does transcript supporting evidence work with this change ?

Zhicheng-Liu:
Hi @ens-st3, not sure about what you mean and how to test it. But this change will only affect the $gene variable at this line (https://github.com/Ensembl/ensembl-webcode/blob/release/87/modules/EnsEMBL/Draw/GlyphSet_transcript.pm#L1461). It will only affect items that have colour_key such as [gene.logic_name]_[gene.biotype]. Maybe @ens-ap5 will know?

ens-st3:
Won't it also potentially affect $colour (https://github.com/Ensembl/ensembl-webcode/blob/release/87/modules/EnsEMBL/Draw/GlyphSet/TSE_transcript.pm#L43) ?

TSE_transcript is used for supporting evidence, for example:
http://www.ensembl.org/Homo_sapiens/Transcript/SupportingEvidence?db=core;g=ENSG00000128573;r=7:114414997-114693768;t=ENST00000403559

I can't find an example in plants though - can you check if if you have any ? The data for these comes from the supporting_feature and transcript_supporting_feature tables.

Zhicheng-Liu:
I checked Metazoa and found the changes works for this page: http://metazoa.ensembl.org/Anopheles_gambiae/Transcript/SupportingEvidence?g=AGAP004707;r=2L:2358158-2431617;t=AGAP004707-RA

ens-st3:
Confirmed OK with Ensembl as well. Thx

PR #525, opened 2017-01-06
Move minify types list into SiteDefs to make it configurable

nicklangridge:


nicklangridge:
Before merging this please see comments here https://www.ebi.ac.uk/panda/jira/browse/ENSEMBL-4782?filter=11757

nicklangridge:
Need to think of better way to implement this in context of existing flags for controlling minification

PR #530, opened 2017-02-08
fix: check if a config exists before dereferencing it

Zhicheng-Liu:
Unfortunately for Plants compara, there are other species from other divisions.
But we do not have species defs for those species.

Fixes ENSEMBL-4853.


Zhicheng-Liu:
Could someone please merge the change in postreleasefix/87 to release/87 and master branches please? Many thanks!

ens-hsr:
Done

PR #537, opened 2017-02-22
Add exporting web root to startup command

Zhicheng-Liu:
This is to allow using this environment variable to "Include" ensembl-webcode's httpd.conf
in plugins' httpd.conf.


Zhicheng-Liu:
We will add exporting the environment in the start hook inside eg-web-common. Closing this for now.

PR #540, opened 2017-03-03
ENSEMBL-4875 - Masthead issues on search page

sanjay-eg:


ens-ap5:
Hi Sanjay

I've come up with a more elegant solution than regexing the object type. There's now a method 'has_tabs' in Configuration which returns 1 or 0, and is used by the template code to set the masthead height. I think everything should just work now for both e! and EG.

PR #541, opened 2017-03-05
Use new Compara methods to get species/clade names

muffato:
From e88 we have a number of new methods to get common names (also known as "ensembl alias" or "display name"), scientific names, etc. This simplifies several modules, and removes the need of building lookup tables in the ConfigPacker (which @Zhicheng-Liu should like). 

I don't know how to test the changes at the EBI, so I have used my Sanger sandbox with the e87 databases instead.

Pages affected are:
 /Homo_sapiens/Gene/Compara_Tree?g=ENSG00000139618 : common names on the collapsed nodes, clade names in the zmenus, ability to hide/show some clades (in Configure This Page)
/Homo_sapiens/Gene/Compara_Paralog?g=ENSG00000189167 : the "Ancestral taxonomy" column
/info/genome/compara/prot_tree_stats.html?page=coverage and the other pages in the same section : Species/clade names

### DIFF
--- a/modules/EnsEMBL/Draw/GlyphSet/genetree.pm
+++ b/modules/EnsEMBL/Draw/GlyphSet/genetree.pm
@@ -674,12 +674,10 @@ sub features {
     my %genes;
     my %leaves;
 
-    my $species_tree_node_id = $tree->get_tagvalue('species_tree_node_id');
+    my $species_tree_node = $tree->species_tree_node();
     my $node_name;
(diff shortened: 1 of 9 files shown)


muffato:
Added an info panel to link out to the HAL installation procedure
/info/genome/compara/mlss.html?mlss=835

muffato:
Added another commit that fixes ENSWEB-2818

PR #550, opened 2017-03-24
Minor fix for variant URL in the table + remove useless module

ens-lgil:
The module LocationsNewTable.pm was added by mistake (it was a test module).


ens-ds23:
Thanks!

PR #552, opened 2017-03-31
Display the gene stable ID if there is no gene symbol available

ens-lgil:


ens-lgil:
Hi,
Would it be possible to push this fix for e!89. please ?
Thanks,
Laurent

ens-lgil:
Thank you Harpreet!

PR #555, opened 2017-04-21
Update for release/89 api changes.

mn1:


mn1:
The arrays in the first loop are different from the arrays in the second loop. So the keys should not be mismatched.

The hash should be the same as what you are getting right now. In your current code the first loop does the check:

https://github.com/Ensembl/ensembl-webcode/blob/release/87/modules/EnsEMBL/Web/Object/Transcript.pm#L1104

and the second doesn't:

https://github.com/Ensembl/ensembl-webcode/blob/release/87/modules/EnsEMBL/Web/Object/Transcript.pm#L1130

The code I provided should be building the same hash.

ens-ap5:
OK, no worries - just wanted to be sure it wasn't an error!

PR #561, opened 2017-05-15
Move the phenotype link from the column 'Associate loci' to the colum…

ens-lgil:
…n 'Phenotype, disease and trait' and remove the column 'Associate loci'


ens-lgil:
Hi,
Would it be possible to add this change for e!90, please ?
Thanks,

PR #566, opened 2017-06-02
Use PRODUCTION_NAMES

Zhicheng-Liu:


ens-ap5:
Hi Zhicheng
If you want the list of names to be the same as before (i.e. ucfirst), you should change it to $species_defs->multi_hash->{'ENSEMBL_DATASETS'}.

PR #568, opened 2017-06-27
fix for ENSREGULATION-774 jira issue

ilavidas:
Hopefully this should fix this issue https://www.ebi.ac.uk/panda/jira/browse/ENSREGULATION-774
Please use the latest master branch from funcgen's repository.


ens-hsr:
This should be on staging now - can you please confirm if it's fixed?

PR #573, opened 2017-07-17
Fix datatables nested form issue

nicklangridge:
Update the datatables code to use div elements instead of form elements for storing configuration params. Form elements cannot be nested and so previously it was not possible to use a datatable inside a form.
See also ENSWEB-2231
Based on the VectorBase modification here https://github.com/EnsemblGenomes/eg-web-vectorbase/commit/28acd78ce41f2110564e5c7efbb9bca7c53336fa


ens-ap5:
Merging this on my sandbox, as it needs some tweaks for compatibility with current webcode.

PR #577, opened 2017-08-07
Changes to the sql as discussed with @ens-hsr

mn1:


ens-st3:
Can you remind me what this is for ?

mn1:
I don't even remember this, but by looking at the sql:

Probe and probe set names are not unique. Especially, if there is more than one version of the same array, a probe or probe set may have the same name on each version. The new sql takes this into account, by not only aggregating by name, but also by the name of the probe or probe set id.

ens-st3:
I don't think this will work since the unique ID for the Solr document is probe_name. If we have multiple records with the same probe_names in the XML which I think this will do, then I think only the last will end up in Solr. The way to do it I think would be to put the probe_names in the description.

Can you send me an example so I can dig in a bit more ?

mn1:
Probes sharing the same name can be found like this:

```
r1 homo_sapiens_funcgen_91_38 -e "select name, count(probe_id) as c from probe group by name having c > 1 limit 10"

name    c
0       1575
10      3
100     2
1000    2
10000   2
100000  3
1000000 2
1000001 3
1000002 2
1000003 3
```

ens-st3:
I'll merge it and see how it works

PR #582, opened 2017-08-17
Don't delete config if key wasn't renamed

nicklangridge:
When renaming keys, allow for the case where production_name and species.url are the same.


nicklangridge:
Closing as have submitted cleaner solution

PR #584, opened 2017-08-21
Remove WTSI references

andrewyatz:
Issue originally raised in ENSWEB-4136. We should get this out before 90 if possible


ens-ap5:
Will do - though I think we can just remove that DAS file!

PR #586, opened 2017-09-20
Use the NewTable lib and fix some missing data for the multiallelic v…

ens-lgil:
…ariants


ens-lgil:
This should make the Protein Variants page faster and more interactive with the different filters:
http://www.ensembl.org/Homo_sapiens/Transcript/ProtVariations?t=ENST00000393330

ens-st3:
@jyothishnt - can you take a look at this when time allows ?

PR #594, opened 2017-11-23
Updates to make the source views work with the new api

mn1:


mn1:
P.S.: This is about:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4267

ens-ap5:
Code looks reasonable to me!

PR #603, opened 2018-01-03
Fix minor issue with empty object

ens-lgil:
If possible we would like to push this fix to master as well

### DIFF
--- a/modules/EnsEMBL/Web/Component/StructuralVariation/PopulationFrequency.pm
+++ b/modules/EnsEMBL/Web/Component/StructuralVariation/PopulationFrequency.pm
@@ -92,6 +92,7 @@ sub table_data {
 
   # Get population structure
   foreach my $svpf (@$data) {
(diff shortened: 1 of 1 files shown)


ens-ap5:
Hi Laurent - To get the fix onto master you'd need to make the pull request against postreleasefix/91, but it doesn't look like it's been created yet.

ens-lgil:
@ens-ap5 should I create the branch on my repo then ?

ens-ap5:
No - we produce it automatically, I think, using one of our git hooks. I’ll ask the new paladins when they get back from lunch, since Harpreet is no longer in charge!

Anne

ens-ap5:
As we don't have the postreleasefix branch in place yet and this is a simple commit, we'll merge it into release/91 and manually cherry-pick it onto master.

ens-lgil:
Thank you Anne!

PR #608, opened 2018-02-06
Link to pathway diagram for InterPro-derived KEGG xrefs

james-monkeyshines:
This is the configuration that EG have had for some time. The xref here is actually a compound of two (or more...) things. Firstly a pathway, and then a set of enzymes (EC IDs) - usually it's just one enzyme, but not always. So, rather than linking to a single enzyme page, we can link to a pathway page, with all the enzymes highlighted in red.


ens-st3:
Thanks

PR #612, opened 2018-02-16
Fix for LDPlot redirecting to LD after selecting populations

imran-ebi:
This fix is for ENSWEB-4360. The sandbox link is: http://ves-hx-78.ebi.ac.uk:8300/Homo_sapiens/Variation/LDPlot?db=core;pop1=373508;r=9:22125004-22126004;v=rs1333049;vdb=variation;vf=906984.

### DIFF
--- a/modules/EnsEMBL/Web/Component/PopulationSelector.pm
+++ b/modules/EnsEMBL/Web/Component/PopulationSelector.pm
@@ -25,15 +25,27 @@ use base qw(EnsEMBL::Web::Component::MultiSelector);
 
 sub _init {
   my $self = shift;
- 
+  my $action = '';
+
   $self->SUPER::_init;
- 
+
   $self->{'link_text'}       = 'Select populations';
   $self->{'included_header'} = 'Selected Populations';
   $self->{'excluded_header'} = 'Unselected Populations';
   $self->{'url_param'}       = 'pop';
   $self->{'rel'}             = 'modal_select_populations';
-  $self->{'url'}             = $self->hub->url({action => 'LDPlot'}, 1);
+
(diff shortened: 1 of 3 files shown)


ens-ma7:
@imran-ebi this change will probably break other places where the population selector is used ( for example EnsEMBL::Web::Component::Location::SelectPopulation). You can see if the following page still work for you:
http://www.ensembl.org/Homo_sapiens/Location/LD?db=core;focus=variation;pop1=373508;r=9:22115504-22135503;v=rs1333049;vdb=variation;vf=906984

maybe have an if statement to check if component is LDPlot then URL is LDPlot else LD

imran-ebi:
@ens-ma7 Thanks for mentioning it. It did break when I checked in the LD page. I will put a check in there to insert the correct page as the action.

imran-ebi:
@ens-ma7 I have added the check in the latest commit

ens-ma7:
@imran-ebi the way you done it is ok but a bit long, just see if you can use $hub->action to find out what page you are currently on (i just remember it might work and might not work because of the configuration panel) and then use the action as the url.

imran-ebi:
@ens-ma7 Your suggestions was great. The fix is now only a one liner and doesn't have any loops.

ens-ma7:
@imran-ebi great the suggestion work but you left one unused variable in the commit ($action), can you remove it please.

imran-ebi:
@ens-ma7 Sorry about that. I totally missed that one 😞 I have removed it now.

imran-ebi:
@ens-ma7 There was another issue identified by @jyothishnt. Have pushed the fixed for that as well.

PR #613, opened 2018-02-19
Add JS library to highlight the Shell code in documentation (e.g. VEP…

ens-lgil:
… script documentation) like we already have for Perl code (92_highlight_perl.js)


ens-st3:
Can you take a look @ens-ma7 ?

ens-ma7:
Hi @ens-lgil can you please include your sandbox URL? And where is this JS file used?

ens-lgil:
Hi @ens-ma7 ,
Sure: the JS file is used in the VEP script documentation portal:
http://ves-hx2-76.ebi.ac.uk:5060/info/docs/tools/vep/script/index.html
(basically all the HTML files under /info/docs/tools/vep/script/)
However my sandbox is (painfully) very slow ...

ens-ma7:
Thanks @ens-lgil , is that JS library only used for the vep documentation?  I am a bit cautious that it goes in the main htdocs, i will suggest to put this in the public-plugins documentation section where the VEP html file is. We can probably have a chat about this.

ens-lgil:
Yes, that was my first thought: put the JS script in public_plugins.
The other thing is that the JS scripts concerning the Perl highlighting is in ensembl-webcode and is only used in the API tutorials (in public_plugins).
Maybe we could move them to public_plugins to have all the highlighting code at the same place.

ens-ma7:
There are a few more files related to the highlighting:
92_highlight.css
92_highlight_main.js
92_highlight_perl.js
93_highlight_start.js (this is the main one that initialise all highlighting)

And they are quite old (2014) and all of them have been added by you except the 93 which was by Dan. Can you confirm if they are all related to the documentation highlighting, if that's the case we can move it to public-plugins.

ens-lgil:
Yes they are all related to the documentation highlighting.

ens-ma7:
Ok i am going to accept the PR and then move all the highlighting files to public-plugins/docs/htdocs.

PR #615, opened 2018-02-21
Display population structures that only have 2 layers

at7:
Display populations that only have 2 layers: super_population -> sub_population. At the moment we only deal with 3 layers (1000 genomes project) or 1 layer (everything else). 
I test views for human and sheep and all is working fine. But getting the PR in early would give us more time for testing. Probably @ens-lgil knows the code best and can comment.

### DIFF
--- a/modules/EnsEMBL/Web/Component/Variation/PopulationGenotypes.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/PopulationGenotypes.pm
@@ -253,9 +253,19 @@ sub format_table {
   push @ids, $all if $all;
   my @super_order = sort {$tree->{$a}{'name'} cmp $tree->{$b}{'name'}} keys (%$tree);
   foreach my $super (@super_order) {
-    next if ($all && $super == $all); # Skip the 3 layers structure, which leads to duplicated rows
-    push @ids, $super;
     my $children = $tree->{$super}{'children'} || {};
+    my $next_level_has_children = 0;
+    foreach my $child_id (keys %$children) {
+      if ($tree->{$child_id}{'children'}) {
+        $next_level_has_children = 1;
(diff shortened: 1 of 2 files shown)


modules/EnsEMBL/Web/Component/Variation/PopulationGraphs.pm
```diff
@@ -81,7 +81,12 @@ sub content {
   }
   
   # Create graphs
+  my $population_count = scalar keys %$pop_freq;
+  my $too_many_populations = ($population_count > 35);
```
ens-lgil: Might be good to put the "35" value in a variable, for future maintenance

modules/EnsEMBL/Web/Component/Variation/PopulationGenotypes.pm
```diff
@@ -253,9 +253,19 @@ sub format_table {
...
-    next if ($all && $super == $all); # Skip the 3 layers structure, which leads to duplicated rows
-    push @ids, $super;
     my $children = $tree->{$super}{'children'} || {};
+    my $next_level_has_children = 0;
+    foreach my $child_id (keys %$children) {
+      if ($tree->{$child_id}{'children'}) {
+        $next_level_has_children = 1;
+      }
+    }
+    next if ($all && $super == $all && $next_level_has_children); # Skip the 3 layers structure, which leads to duplicated rows
```
ens-lgil: if/else a bit complicated:

if ($all && $super == $all) {
  next if ($next_level_has_children); # Skip the 3 layers structure, which leads to duplicated rows
} else {
  push @ids, $super;
}

at7:
Thank you for the comments. I updated the code accordingly.

PR #617, opened 2018-02-22
bugfix/improvement: there is a cleaner way of checking if the species is "low-coverage" or not

muffato:
Seems to be fixing ENSWEB-4401

Tested on my sandbox: http://ves-hx2-76.ebi.ac.uk:5080/Capra_hircus/Location/Compara_Alignments?align=1136;db=core;r=5:23186176-23233107

[cc @carlacummins ]


muffato:
Thank you Anne !

PR #619, opened 2018-02-23
New way of listing alignments

muffato:
[ cc @ens-emily ]

I have copied the [EG way](http://fungi.ensembl.org/compara_analyses.html) of [listing the pairwise alignments](http://staging.ensembl.org/info/genome/compara/analyses.html) (see [my sandbox](http://ves-hx2-76.ebi.ac.uk:5080/info/genome/compara/analyses.html)) with two changes:
 - added the common names of species
 - replaced the capital letter codes LASTZ_NET, SYNTENY, etc with capitalized text 

There is still an open question. In Ensembl we were sorting the species according to the phylogeny, orienting the tree with the favourite species (by default: human, mouse, zebrafish). Other options are sorting by common name or by scientific name.

### DIFF
--- a/modules/EnsEMBL/Web/Document/HTML/Compara.pm
+++ b/modules/EnsEMBL/Web/Document/HTML/Compara.pm
@@ -42,18 +42,6 @@ sub common_name {
   return $self->hub->species_defs->get_config($name, 'SPECIES_COMMON_NAME');
 }
 
-sub get_genome_db {
-  my ($self, $adaptor, $short_name) = @_;
-
-  my $all_genome_dbs = $adaptor->fetch_all;
-  $short_name =~ tr/\.//d;
-  foreach my $genome_db (@$all_genome_dbs) {
-    if ($genome_db->get_short_name eq $short_name) {
-      return $genome_db;
-    }
-  }
-}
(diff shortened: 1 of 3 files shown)


ens-ma7:
@muffato i cant see where the two new functions are being used (draw_pairwise_alignment_list  and the other one).

muffato:
Right, that would be on public-plugins

```diff
diff --git a/ensembl/htdocs/info/genome/compara/analyses.html b/ensembl/htdocs/info/genome/compara/analyses.html
index 6994ca3..687a143 100644
--- a/ensembl/htdocs/info/genome/compara/analyses.html
+++ b/ensembl/htdocs/info/genome/compara/analyses.html
@@ -23,7 +23,7 @@
 
 <h2>Alignments available</h2>
 
-[[SCRIPT::EnsEMBL::Web::Document::HTML::Compara::BlastZ]]
+[[SCRIPT::EnsEMBL::Web::Document::HTML::Compara::draw_pairwise_alignment_list()]]
 
 </body>
 </html>
```

muffato:
Thank you Ridwan
I have pushed the related commit to public-plugins

PR #632, opened 2018-05-31
Display extra information from ClinVar in the phenotype-related pages

ens-lgil:
- Distinct source entry for each submission
- List of PubMed IDs as evidence


ens-lgil:
Hi,
We would like to see these updates in the release `93`.
Please, could you have a look at it ?
Thanks,
Laurent

PR #636, opened 2018-06-09
Multiple-alignment stats

muffato:
The most important change is that we store the alignment statistics in a different object. This fixes the stats page: http://staging.ensembl.org/info/genome/compara/mlss.html?mlss=1142
See: http://http://ves-hx2-77.ebi.ac.uk:5083/info/genome/compara/mlss.html?mlss=1142

The other change is to use the pre-built species-tree rather than creating a new one


ens-ds23:
I can't verify this on my sandbox, but that's probably because of my confusion over MLSSids vs release versions. I'll accept the PR anyway and we can check it on staging as it only affects doc pages.

muffato:
Thanks @ens-ds23 . I can show you on my sandbox if you want

PR #637, opened 2018-06-09
bugfix/improvement: cleaner way of checking if the species is "low-coverage" or not

muffato:
This is a follow-up to 0d885cf7767bc7c85f0f8b61581511d6c832d0ed


ens-ds23:
Confirmed on sandbox.

PR #647, opened 2018-09-04
Default to NULL instead of empty string when no species code

nicklangridge:
Species code is not compulsory, but the unique constraint on the species db table will not allow duplicate rows with empty string for this field.


ens-ap5:
I think we can do this fix for now, and do the improvements Steve mentioned when we can. I'll raise a JIRA ticket for that.

PR #649, opened 2018-09-06
Fix the 'hidden_columns' Table option and fix invisible top horizonta…

ens-lgil:
…l scrollbar on Chrome in VEP result page


ens-st3:
Can you take a look @ens-ma7 ?

PR #659, opened 2018-10-26
Removed the option to hide so-called "low-coverage" species

muffato:
The information is not provided any more on recent Vertebrates assemblies. I think it's (almost) never been set in EG either.
:warning: I haven't tested this on my sandbox !

Then you can also remove the line in public-plugins/ensembl/conf/ini-files/DEFAULTS.ini that has "LOWCOVERAGE"


imran-ebi:
Thanks for the pull request @muffato. I have merged it now.

PR #664, opened 2018-11-14
Show the variant translation 'end' coordinates when the 'start' and '…

ens-lgil:
…end' are different (deletion or substitution overlapping more thant 1 amino acid)

## Description

Show the variant translation `end` coordinates when the `start` and `end` are different (deletion or substitution overlapping more thant 1 amino acid).
By default the table is only displaying the `start` coordinate.

## Views affected

- Gene - Variant table
- Transcript - Variant table


ens-st3:
Can you send a sandbox link ?

ens-lgil:
Here is on my sandbox:
http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Transcript/Variation_Transcript/Table?g=ENSG00000185198;r=19:685546-695498;t=ENST00000613411
The examples display when you select the consequences `frameshift_variant` and `inframe_insertion`

PR #667, opened 2018-11-26
Feature/pdb viewer

ens-lgil:
## Description

Code to display PDBe 3D viewer in Ensembl (most of the code is in `public-plugins`).
Associated with the `public-plugins` PR https://github.com/Ensembl/public-plugins/pull/165

## Views affected

There will be a new view for Human in the `Transcript` and `Variation` portals, e.g.:
- Transcript: http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Transcript/PDB?t=ENST00000304043
- Variation: http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Variation/PDB?v=rs699


modules/EnsEMBL/Web/Component/Variation/Explore.pm
```diff
@@ -97,6 +101,10 @@ sub content {
     {'title' => 'Citations',                                          'img' => '96/var_citations.png',              'url' => $cit_url,      'count' => $cit_count   },
   );
 
+  if ($self->hub->species eq 'Homo_sapiens') {
```
ens-ma7: I think for now this is fine but what if other species have this enable? Or will it just be human? One way of doing it is having a flag in each species ini file, this way to enabling it for another species will just be adding the flag in the INI file, its not critical for now if it is going to be only human but something to remember for the future

modules/EnsEMBL/Web/Configuration/Variation.pm
```diff
@@ -124,6 +124,13 @@ sub populate_tree {
     { 'availability' => 'variation has_ldpops has_samples not_somatic', 'no_menu_entry' => 1 }
   );
 
+  if ($self->hub->species eq 'Homo_sapiens') {
```
ens-ma7: i dont quite understand the if check for human, does the availability check (is_coding) not make sure of this. Does is_coding return something for other species?

ens-lgil: We only have this mapping for Human so we don't want to show this icon/entry for non human species.
The `is_coding` only checks if the variant overlaps a coding exon (this could be used for other views in the future).

PR #684, opened 2019-02-22
check if there is a vcf config file first

at7:
## Requirements

None

## Description

This is a fix for a bug which I noticed on [staging](http://staging.ensembl.org/Meleagris_gallopavo/Variation/Explore?r=18:72663-73663;v=rs119118426;vdb=variation;vf=5810). The problem being that there was no vcf config file under public-plugins/ensembl/conf/json/ for the species.

## Views affected

Variant summary page

## Possible complications

Is $c->{'CONFIG'} always pointing to public-plugins/ensembl/conf/json/ if there is a config file for the given species stored in that directory?

## Merge conflicts

None

None

### DIFF
--- a/modules/EnsEMBL/Web/Component/Variation/Summary.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/Summary.pm
@@ -725,6 +725,9 @@ sub clinical_significance {
 
 sub change_tolerance {
   my $self = shift;
+  my $adaptor = $self->hub->get_adaptor('get_VariationFeatureAdaptor', 'variation');
+  my $c = $self->hub->species_defs->ENSEMBL_VCF_COLLECTIONS;
+  return unless (defined $c->{'CONFIG'});
   my $object = $self->object;
   my ($CADD_scores, $CADD_source) = @{$object->CADD_score};
   my ($GERP_score, $GERP_source) = @{$object->GERP_score};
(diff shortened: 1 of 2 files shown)


ens-ap5:
This looks OK, but do you know why we have a flag in the config called 'ENABLED'? If it's significant, it would better to check for this value, but as we always set it to 1 it seems a bit pointless. See here:
https://github.com/Ensembl/ensembl-webcode/blob/release/95/modules/EnsEMBL/Web/SpeciesDefs.pm#L575

at7:
It looks a bit more complicated to get filename and confdir information where I want to check for existence of a vcf config file. Enabled differentiates between different values: 0 (no vcf), 1 (only vcf), 2 (vcf and database). This is how the variation API uses the value stored in the use_vcf method.

ens-ap5:
Oh, OK - anything to keep the variation API happy! However looking at it more closely, I think it would be better practice to put this check in the methods that return data, to keep the code as cleanly MVC as possible. Thus in Variation::Object::CADD_score, just add:

return unless $self->hub->species_defs->ENSEMBL_VCF_COLLECTIONS;

at the start of the method. I also note that the CADD_score method can return an array of undefs, which is not what the change_tolerance method is checking for, here:

$self->hub->species_defs->ENSEMBL_VCF_COLLECTIONS;

Feel free to come and talk to me if this doesn't make sense.

at7:
Thank you for the feedback. Did I get it right or where you thinking of something else?

ens-ap5:
No, you're right - I needed to get some coffee and reread Component::Variation::Summary :)

at7:
Thank you!

PR #690, opened 2019-04-15
[ENSWEB-4928] do not overwrite a section with defaults if it is already in the hash

jyothishnt:
## Requirements

## Description

DEFAULT_XREFS entries in homo_sapiens.ini file were missing from the gene external refs table. 
There is a problem in parsing ini files. As far as I understand we should not overwrite a section with the default values if the tree already has that section added.

## Views affected

Gene External references table and Transcripts table in Gene Summary view. But may be other views as well.

## Possible complications

This change may affect other views as well which I am not aware of, as it will make changes to config.packed file. Until now though the ini files were read properly, the db_tree was not properly generated. So you may see some changes in other views too.

## Merge conflicts

No conflicts. Please use mpush after merging.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4928

### DIFF
--- a/modules/EnsEMBL/Web/SpeciesDefs.pm
+++ b/modules/EnsEMBL/Web/SpeciesDefs.pm
@@ -543,7 +543,7 @@ sub _read_in_ini_file {
           $tree->{$current_section} ||= {}; # create new element if required
           
           # add settings from default
-          if (defined $defaults->{$current_section}) {
+          if (!%{$tree->{$current_section}} && defined $defaults->{$current_section}) {
             my %hash = %{$defaults->{$current_section}};
             
             $tree->{$current_section}{$_} = $defaults->{$current_section}{$_} for keys %hash;


ens-ap5:
So, we're assuming that 1) if we have settings already, they're all present and correct, and 2) if we don't, we populate them from defaults. This could potentially break things if assumption 1 is incorrect and we've been relying on the default settings to fill in the gaps - but if we only populate the keys that _aren't_ in the current section, that could break things too by giving us a weird hybrid config! I would prefer to go with the hack that Steve mentions, and make this change in release 97 when we have more lead time to spot bugs.

ens-ma7:
I agree with @ens-ap5 , its a bit scary when the first if condition is true for others. It will be best to warn the tree and see whats being populated and check each values and then test other views. Probably best to go with Steve hack.

jyothishnt:
OK. We will discuss this in detail once I am back from holidays.

ens-ma7:
The change is fine but it needs testing, it was a bit scary for a hotfix. Best to push this to master and then test.

ens-st3:
not for release 96!

ens-ap5:
I agree with the above comments - put it onto **master** so we can test it thoroughly, not onto postreleasefix/96, and then I'm happy to merge it.

jyothishnt:
Thanks all. I will merge it now.

PR #692, opened 2019-05-02
Be more careful as not all genomes are in the compara db

muffato:
## Description

This commit allows viewing the gene-tree page even if the Compara database doesn't have all the genomes.
I did that change locally a while ago, after trying to visualize our production gene-tree databases (which are missing either the mouse-strains or the non-murinae species) on my sandbox. Yesterday, @jyothishnt noticed that the same error happens on staging as the database is the e96 one (thus missing the e97 species). See http://staging.ensembl.org/Homo_sapiens/Gene/Compara_Tree?g=ENSG00000139618;r=13:32315086-32400266

## Views affected

Gene-tree view. Fixed on my sandbox: http://ves-hx2-77.ebi.ac.uk:5081/Homo_sapiens/Gene/Compara_Tree?g=ENSG00000139618;r=13:32315086-32400266 (note that the sandbox is using the same database as staging.ensemblorg)

## Possible complications

As the page will now be displayed on staging, people may think that the new Compara database is available

## Merge conflicts

This commit is on top of master. There shouldn't be any conflict.

N/A


muffato:
As I am doing some tests before the handover, my sandbox is now pointing at our new e97 databases

PR #694, opened 2019-05-07
[ENSWEB-4912] Use data from Ontology Lookup Service for the biotypes page

azangru:
## Description

This PR replaces static html content of the [biotypes page](http://www.ensembl.org/info/genome/genebuild/biotypes.html) with the definitions retrieved from Ontology Lookup Service. The list of definitions is retrieved once, at the build time of the MULTI package, and then used when rendering the page. This is done in a similar way to how the [glossary page](http://www.ensembl.org/info/website/glossary.html) is displayed.

See related PR in the public-plugins repo: https://github.com/Ensembl/public-plugins/pull/208

## Views affected

[Biotypes page](http://www.ensembl.org/info/genome/genebuild/biotypes.html).

Screenshot of the updated page:
image

## Possible complications

None that I am aware of.

https://www.ebi.ac.uk/panda/jira/projects/ENSWEB/issues/ENSWEB-4912

### DIFF
--- a/modules/EnsEMBL/Web/Component/Help/Biotypes.pm
+++ b/modules/EnsEMBL/Web/Component/Help/Biotypes.pm
@@ -1,0 +1,73 @@
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2018] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package EnsEMBL::Web::Component::Help::Biotypes;
+
+use strict;
+use warnings;
+no warnings "uninitialized";
+
+use base qw(EnsEMBL::Web::Component::Help);
+
+sub _init {
+#  my $self = shift;
+#  $self->cacheable( 0 );
+#  $self->ajaxable(  0 );
+#  $self->configurable( 0 );
+}
+
+sub content {
+  my $self = shift;
+  my $hub = $self->hub;
+  my %biotypes = $hub->species_defs->multiX('ENSEMBL_BIOTYPES');
+  my $rendered_term_tree = render_term(\%biotypes);
+
+  return "<ul>${rendered_term_tree}</ul>"
+}
+
+sub render_term {
+  my $term = shift;
+
+  my $label = $term->{label};
+  my $description = join ' ', @{$term->{description}}; # description field is an array;
+  my $rendered_children = '';
+  my $children = $term->{children};
+
+  if ($children and scalar(@{$children})) {
+    $rendered_children = render_children($children);
+  }
+ 
+  my $html = "<li><strong>${label}:</strong> ${description} ${rendered_children}</li>";
+  return $html;
+}
+
+sub render_children {
+  my $children = shift;
+  my $rendered_children = '';
+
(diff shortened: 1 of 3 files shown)


modules/EnsEMBL/Web/Component/Help/Biotypes.pm
```diff
@@ -0,0 +1,73 @@
...
+use warnings;
+no warnings "uninitialized";
+
+use base qw(EnsEMBL::Web::Component::Help);
+
+sub _init {
+#  my $self = shift;
+#  $self->cacheable( 0 );
+#  $self->ajaxable(  0 );
+#  $self->configurable( 0 );
```
azangru: Not sure if any of this is relevant or whether this can be safely removed.

ens-ap5: I don't know that all of it is relevant, but it won't break anything if you uncomment it. I think, though, that you might want to set 'cacheable' to 1 since the page isn't going to change much between releases.

ens-ap5: Also, just as a coding style tip, you don't need the curly braces around interpolated scalar variables, eg. in lines 41 and 56 below. However they are recommended when dereferencing variables, mainly so that you can write e.g. @{$children||[]} (line 52) - because Perl is untyped, you can't rely on $children to be an array reference (even if you wrote the code that creates $children, because someone else could modify it later and screw things up!). And since the most likely "wrong" return value is undef, this "or []" construction prevents ugly code barfs.

modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -1034,7 +1035,45 @@ sub _summarise_website_db {
...
+    # NOTE: a term url passed as a parameter to the terms endpoint
+    # should use http, lack www, and be twice(!) url-encoded
+    my $biotype_url = 'http://ensembl.org/glossary/ENSGLOSSARY_0000025';
+    my $safe_biotype_url = uri_escape(uri_escape($biotype_url));
+    my $biotype_term_url = $terms_endpoint . '/' . $safe_biotype_url;
+ 
+    sub fetch_term {
+      my %result;
+      my $url = $_[0];
+      my $term = $self->_get_rest_data($url);
```
azangru: When restarting the server, I am seeing the message: `Variable "$self" will not stay shared`. Not sure what that means; but most likely that's because I declare a subroutine within a subroutine. Don't know what the proper course of actions is (move the `fetch_term` and `fetch_children_terms` functions to the top level? pass `$self` to them as an argument?)

modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -1034,7 +1035,45 @@ sub _summarise_website_db {
...
+      $result{synonyms} = $term->{synonyms};
+
+      if ($children_link) {
+        $result{'children'} = [ fetch_children_terms($children_link) ];
+      }
+      return \%result;
+    }
+
+    sub fetch_children_terms {
+      my $url = @_[0];
```
azangru: When restarting the server, perl compiler opines that "Scalar value @_[0] better written as $_[0]". I've almost given up all hope to understand perl's type annotations. I thought, `@_[0]` is pretty idiomatic. Just as `shift` is; only perhaps more explicit. Do we have any conventions regarding how to write this thing?

ens-ap5: Yes. You only ever use @ and % to refer to the entire array/hash (and when dereferencing). References, array slices, hash elements, etc always use $.

modules/EnsEMBL/Web/ConfigPacker.pm
ens-ap5: I've never seen that error message before! Methods within an object should always be called on $self, ($self->fetch_term) and then you write "my $self = shift" as the first line of the method (or, for example, "my ($self, $arg) = @_;" if passing an argument).

azangru:
@ens-ap5 @jyothishnt I have updated the code since the latest review; could you have a quick glance at it and say whether you are happy for it to be merged? I was worried about the general content of the page (it became a bit less compared to the static html version), but @ens-st3 says he thinks it's fine; so I guess it's only the code quality that we are currently after.

ens-st3:
Go ahead and merge Andrey (no need for review from Dan)

PR #701, opened 2019-05-22
Hide phenotypes associated with non mapped variants

ens-lgil:
## Description

Minor fixes to hide phenotype entries which are associated with variants that no longer map to the genome. 

## Views affected

The Variation/Phenotype pages.
This cause an issue because the Phenotype page is made available, but when you click on it, it's empty (mainly because the page is looking at the variation_feature, in case a variant maps to several locations).

Here is an example on my sandbox:
http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Variation/Explore?db=core;v=rs505922;vdb=variation
versus on the live website:
http://www.ensembl.org/Homo_sapiens/Variation/Explore?db=core;v=rs505922;vdb=variation

## Possible complications

We are going to cleanup the phenotype_feature table in e98 to avoid this kind of issue in the future.

### DIFF
--- a/modules/EnsEMBL/Web/Component/Variation/Explore.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/Explore.pm
@@ -69,7 +69,7 @@ sub content {
     }
   }
 
-  if ($avail->{'has_ega'}) {
+  if ($avail->{'has_ega'} && $avail->{'has_locations'}) {
     $pheno_url   = $hub->url({'action' => 'Phenotype'});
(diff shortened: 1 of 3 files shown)


ens-lgil:
@azangru, ah I forgot to update this one!

The fix grey out the phenotype icon in the bottom panel and also grey out the `Phenotype data` entry in the left menu.
I will send a new commit with the fix for the `About this variant`.
Thanks!

PR #702, opened 2019-05-22
Add longreads submenu for tracks to configuration menu

imran-ebi:
## Description

This PR separates the Long-Seq tracks in `Genes` submenu (under `Genes and transcripts`) in the configuration menu into a separate submenu called `Longreads`.

## Views affected

The configuration menu will be affected by this. Link to sandbox: http://ves-hx-78.ebi.ac.uk:8320/Homo_sapiens/Location/View?db=core;g=ENSG00000139618;r=13:32315086-32400266

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4875

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -34,9 +34,9 @@ use Sys::Hostname::Long;
 
 ###############################################################################
 ## Ensembl Version and release dates (these get updated every release)
-our $ENSEMBL_VERSION        = 96;            # Ensembl release number
-our $ARCHIVE_VERSION        = 'Apr2019';     # Archive site for this version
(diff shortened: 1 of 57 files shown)


ens-ma7:
@imran-ebi i have tried switching on the longreads track but they are not showing in the region in detail view.

imran-ebi:
@ens-ma7 Thanks for noting this. I checked the site again will all the `Long-Seq` tracks on and only some of them appear in the image. This behaviour can be reproduced on the live site too.

imran-ebi:
As we found out offline, the `Long-Seq` tracks appear only in the correct region. The confusions was because its behaviour is different compared to other tracks.

Also, the `longreads` menu was missing in two of the view configurations (multi bottom and gene summary). This is now fixed.

PR #706, opened 2019-06-07
Fix LOVD external links, which were wrong for the variant insertion type

ens-lgil:
## Description

The variant coordinates built for the LOVD external links were wrong for `insertion` variants.

## Views affected

LOVD short variant ZMenus for Human on the ContingViewBottom panel (Region in detail)


ens-lgil:
Also somehow related to this PR: https://github.com/Ensembl/ebi-plugins/pull/16

PR #710, opened 2019-06-20
[ENSWEB-5042] remove db param from url for Tools

jyothishnt:
## Description

When you open a link from search result with url param db=otherfeatures and then go to Tools page, this param is carried across. It has no issues until a blast result link is clicked where db=core is needed.
So better to remove db param if accessing tools page which is not used in tools.

## Views affected
Only url change in Tools pages

## Possible complications

Nothing I am aware of.

## Merge conflicts

No

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5042

### DIFF
--- a/modules/EnsEMBL/Web/Hub.pm
+++ b/modules/EnsEMBL/Web/Hub.pm
@@ -434,7 +434,8 @@ sub url {
   delete $pars{'v'}  if $params->{'vf'};
   delete $pars{'time'};
   delete $pars{'_'};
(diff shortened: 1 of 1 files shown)


ens-ap5:
Seems harmless enough to me - it's only needed by a few views, and I can't think of a tool that would require it to be passed.

jyothishnt:
Thanks @ens-ap5  I will merge it then.

PR #713, opened 2019-07-02
Fix gene families link for non-vertebrates sites

azangru:
## Description
For some genes, the "About this gene" section of gene summary contains a link to protein families.

image

This link is now generated according to the same rule for both vertebrates and non-vertebrates sites; but in fact the rule should be slightly different. As a result, on non-vertebrates pages the link leads to a page with an error:

image

## Views affected
Gene page (gene summary, "about this gene" section)

fix on local sandbox: http://ves-hx2-76.ebi.ac.uk:8410/Magnaporthe_oryzae/Gene/Summary?db=core;g=MGG_01236;r=2:2789925-2792654;t=MGG_01236T0

## Possible complications
This fix was tested on vertebrates and fungi sandboxes. Protists have the same url structure, so should be fixed with the change in this PR. I could not find genes with gene families for plants, bacteria, or metazoa.

GRCh37 plugin [does not seem to overwrite](https://github.com/Ensembl/ebi-plugins/blob/release/96/grch37/conf/SiteDefs.pm) the ENSEMBL_SITETYPE constant; so it shouldn't break.

## Alternative solutions
We could add a special constant in the SiteDefs file, which will contain the appropriate url fragment (Gene/Family vs Gene/Gene_families) for various sites, but that will involve more changes across various repositories; and I am not sure what to call this url fragment; it looks pretty random.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5083

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -114,6 +114,8 @@ our $SYSLOG_COMMAND                   = sub { warn "$_[0]\n"; };  # command/subr
 our $TIDY_USERDB_CONNECTIONS          = 1;      # Clear user/session db connections after request is finished
 our $SERVER_ERRORS_TO_LOGS            = 1;      # Send all server exception stack traces to logs and send a unique error Id on the browser
 our $ENSEMBL_OOB_LIMITS               = {};     # Child process out-of-bounds limits for live server tweaking
+
+our $GENE_FAMILY_ACTION               = 'Family'; # Used to build the link to gene families page
 ###############################################################################
 
 
--- a/modules/EnsEMBL/Web/Component/Shared.pm
+++ b/modules/EnsEMBL/Web/Component/Shared.pm
@@ -498,7 +498,7 @@ sub about_feature {
     
     my $protein_url = $hub->url({
       type   => 'Gene',
-      action => 'Family',
+      action => $SiteDefs::ENSEMBL_SITETYPE eq 'Ensembl' ? 'Family' : 'Gene_families',
       g      => $gene->stable_id
     });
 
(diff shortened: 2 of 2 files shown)


ens-st3:
Do any other sites, eg N and NV archives and AWS mirrors, overwrite ENSEMBL_SITETYPE ? I'm pretty sure the AWS mirrors don't but I'm not sure about archives.

If we have any current examples of overwriting the action using a plugin configuration then I think we should do that since it's cleaner. If not happy to accept this

ens-ap5:
Should this pull request not go on postreleasefix/97? Or is it only for the next release?

azangru:
@ens-ap5 The JIRA task said release 98, so 🤷‍♂

ens-ap5:
OK, no problem - just double-checking that I hadn't missed anything this release!

ens-ap5:
The pre sites override ENSEMBL_SITETYPE, the archives don't (they use ENSEMBL_SUBTYPE, because you could in theory have an archive of any kind of site).

ens-ap5:
I would suggest using a new variable, GENE_FAMILY_ACTION, set to 'Family' in ensembl-webcode/conf/SiteDefs.pm, then override this in eg-web-common. We should however also check with the relevant teams if this is the correct behaviour for all divisions.

jyothishnt:
To explain this I have to use the terms which I may get £1 penalty.
Ensembl Genomes uses  Ensembl webcode overriding its methods in eg_web_common if it is common for all divisions and to individual eg-web-<plants,fungi,...> for division specific modifications.  So it is better not to add conditional statements inside main ensembl webcode, instead override inside eg-web-* as @ens-ap5 s suggestion. I can help you if you have any doubts on this.

nicklangridge:
> it is better not to add conditional statements inside main ensembl webcode, instead override inside eg-web-* 
I'm not sure about this. Every time we override a sub in eg-web-common (or a division plugin) it's a ticking time bomb as it's hard to keep all the plugins up-to-date with the main webcode and bugs creep in as they get out of sync.
I prefer Anne's suggestion to use a config variable which is used in the main webcode. It is then the variable not the sub which will be overrriden in the plugin.

PR #717, opened 2019-08-08
Improve zmenu for protein align features

msmanoj:
## Description

Added `Alignment length` and `%id` to the zmenu of protein align features.

## Views affected
http://ves-hx2-76.ebi.ac.uk:42229/Homo_sapiens/Location/View?r=17:63992802-64038237

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-2615

### DIFF
--- a/modules/EnsEMBL/Web/ZMenu/Genome.pm
+++ b/modules/EnsEMBL/Web/ZMenu/Genome.pm
@@ -92,16 +92,15 @@ sub content {
   my $cigar_string = $features->[0] ? $features->[0]->cigar_string : undef;
 
   if($align_type and $cigar_string){
-    my $alength = new Bio::EnsEMBL::BaseAlignFeature( -align_type => $align_type, -cigar_string => $cigar_string)->alignment_length();
-    if($alength){
-      $self->add_entry({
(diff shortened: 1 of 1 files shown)


ens-ap5:
A couple of improvements spring to mind:

1. I find it's best to wrap API calls in an eval block, because some of the methods will "throw" if they can't find data, which breaks the whole web page. Usually I create the object within an eval block, then check for the object's existence before calling any subsequent methods. Call me paranoid, but edge cases just lead to RT tickets!

2. For the %id field, it might be better to use '-' rather than undef as the fallback value, so that it's clear to the user that the data is unavailable rather than it being a bug.

msmanoj:
@ens-ap5 I have made those two changes. Thanks

ens-ap5:
I can't see the new values on your sandbox now - have you checked out a different branch?

msmanoj:
Yes, sorry! I've put it back now.

ens-ap5:
Looking good! I'll just ping Jyo since he was a requested reviewer, then merge it in.

msmanoj:
Thanks, Anne.

PR #718, opened 2019-08-12
Feature/precache farm

nicklangridge:
## Description

Add a script to generate the precache by executing the jobs on the farm.
This will speed up the precache generation which is currently very slow.

## Views affected

None

## Possible complications

Nothing anticipated but I'm still testing.

## Merge conflicts

No conflicts.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEBREL-1018

### DIFF
--- a/ctrl_scripts/precache-farm.pl
+++ b/ctrl_scripts/precache-farm.pl
@@ -14,8 +14,10 @@ use Parallel::Forker;
 use List::Util qw(shuffle);
 
 my ($list,@subparts);
+my $workers = 10;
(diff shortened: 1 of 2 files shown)


nicklangridge:
I know the commit history is messy - we can squash merge it :)

ens-ds23:
Looks good to me. It looks like a patched version of precache-local.pl, so I don't think there's much that can go wrong subtly that wouldn't also have happened with the local script (unless we are betrayed by a whacky filesystem).

PR #726, opened 2019-09-03
Show 'Download family alignment' button conditionally, using SiteDefs

azangru:
## Description
The Family proteins page for fungi shows "Download family alignment" button:

image

However, the actual attempt to download family alignment results in an error:

image

The reason for this is that we do not provide sequence alignments for non-vertebrates.

## Views affected
Family proteins (e.g., for fungi, http://fungi.ensembl.org/Multi/Family/Details?fm=MF_00505; for vertebrates: http://www.ensembl.org/Multi/Family/Details?fm=PTHR11289_SF0)

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5255

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -116,6 +116,7 @@ our $SERVER_ERRORS_TO_LOGS            = 1;      # Send all server exception stac
 our $ENSEMBL_OOB_LIMITS               = {};     # Child process out-of-bounds limits for live server tweaking
 
 our $GENE_FAMILY_ACTION               = 'Family'; # Used to build the link to gene families page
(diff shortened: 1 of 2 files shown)


ens-ma7:
@azangru the changes looks good to me but do you have a sandbox URL i can have a look at it.

thanks
Ridwan

azangru:
@ens-ma7 here's url to my sandbox with fungi (the only interesting thing here being that the button is no longer showing): http://ves-hx2-76.ebi.ac.uk:8420/Multi/Family/Details?fm=MF_00505

ens-ma7:
Thanks @azangru i just wanted to check other species and other protein views just in case they had another download button somewhere. I did find a download protein link but thats correct. Everything else looks ok. I am happy with the change.

PR #727, opened 2019-09-04
Update regex that guesses species from query in Psychic

azangru:
## Description
Fixes bug that can be reproduced by the following steps:
- go to home page of ensembl site
- from the dropdown box (in the centre of the screen), select a species (doesn't matter which)
- type "amelogenin" in the search box below
Observe that the search results come for panda (because the word "amelogenin" starts with the substring "amel", which is the short code for panda).

One possible solution is to update the regex in the Psychic module that assumes that if the beginning of a query matches any species present in a species hash, then it must be a query regarding this species. The updated regex will assume that if there is a species name in the query, it will be separated from the actual query by a space.

## Views affected
Search page.

## Possible complications
Don't know

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5134

### DIFF
--- a/modules/EnsEMBL/Web/Controller/Psychic.pm
+++ b/modules/EnsEMBL/Web/Controller/Psychic.pm
@@ -98,7 +98,7 @@ sub psychic {
   #if there is a species at the beginning of the query term then make a note in case we trying to jump to another location
   my ($query_species, $query_without_species);
   foreach my $sp (sort keys %sp_hash) {
-    if ( $query =~ /^\Q$sp\E\b/) {
+    if ( $query =~ /^\Q$sp\E\s/) {
       ($query_without_species = $query) =~ s/\Q$sp\E//;
       $query_without_species =~ s/^ //;
       $query_species = $sp;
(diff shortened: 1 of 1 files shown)


nicklangridge:
Just to note, using word boundary matching (`\b`) in the regex will allow matches for punctuation characters (excluding underscores and hyphens) as well as white-space. Is that intended?

azangru:
@nicklangridge I don't really know what the behaviour of the search field is intended to be. I guess we can change `\b` to `\s` or just ` ` (`/^\Q$sp\E /`) if that's more appropriate?

ens-ap5:
Nick's right - we don't want something like "amel.2" to redirect to panda (assuming that could be a valid gene identifier in, say, a non-vertebrate species.

I would prefer changing the \b to \s, as \E isn't even in my Perl pocket reference!

nicklangridge:
I'm not certain either. But `\s` might be the safer option. A contrived example: Imagine we have IDs like `amel;12345`. If a user tried to search for that, the colon would match the `\b`. Admittedly, it's unlikely we have those IDs, but I don't know for sure.

nicklangridge:
`\Q` and `\E` are used correctly there - they ensure that $sp is escaped appropriately for the regex

PR #728, opened 2019-09-05
Rename strain collection to strain group

msmanoj:
## Description
It looks like meta key `STRAIN_COLLECTION` has been renamed to `STRAIN_GROUP`.
The key `STRAIN_COLLECTION` was used in few other files that are not related to the two issues below, but I have replaced it in all places as it doesn't exist anymore.

- Home pages are now not accessible for strains and breeds. The user will be redirected to the strain page for the parent species.
- Hiding the strains and breeds from the species search dropdown from the homepage.

## Views affected
Species dropdown - http://ves-hx2-76.ebi.ac.uk:42229/index.html
Redirection - http://ves-hx2-76.ebi.ac.uk:42229/Mus_musculus_A_J/Info/Index

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5264
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5125


msmanoj:
Hi @ens-ap5, I have confirmed it with genebuilders that this is the expected behaviour.

ens-ap5:
Cool - I'm happy with merging it, unless any of the other reviewers wants to pipe up?

modules/EnsEMBL/Web/Configuration/Info.pm
```diff
@@ -53,8 +53,9 @@ sub populate_tree {
   my %error_messages = EnsEMBL::Web::Constants::ERROR_MESSAGES;
 
   ## Redirect strains to the strain page for the parent species as long as we're not on the parent species
-  if ($self->hub->action ne 'Strains' && $species_defs->STRAIN_COLLECTION && $species_defs->SPECIES_STRAIN !~ /reference/) {
-    my $url = $self->hub->url({'species' => $species_defs->STRAIN_COLLECTION, 'action' => 'Strains'});
+  if ( $self->hub->action ne 'Strain' && $species_defs->STRAIN_GROUP && $species_defs->SPECIES_STRAIN !~ /reference/
```
ens-ma7: The first check (ne 'Strain') instead of ne 'Strains' is that intentional?

ens-ap5: Oh, well spotted! Definitely should be 'Strains'.

msmanoj: That's done. Thanks.

PR #730, opened 2019-09-18
Fix/data table sorting

azangru:
## Description
Fix for incorrect data table ordering.

**Steps to reproduce the bug**
- visit the [Transcript/Domain/Genes page for mouse](http://www.ensembl.org/Mus_musculus/Transcript/Domains/Genes?db=core;domain=IPR026271;g=ENSMUSG00000028591;r=4:100000001-150000000;t=ENSMUST00000030326)
- notice the large data table in the bottom of the page
- notice that the second column (Genome Location) is sortable
- try sorting this column
- notice that the cells in this column get sorted in an irregular way
- notice too that sorting of the cells in this column yields the same result as if the sorting was done using the first column (Gene)

**Cause of the bug**
During initialisation, the dataTable library gets data from the table by simply reading the contents of each cell. In the case, the contents of the second column of the table are links that look as follows:

```
<a href="/Mus_musculus/Location/View?g=ENSMUSG00000079423">Chromosome 5: 95510758</a>
```

When we try to order rows by values of the second column, the dataTable library performs string comparison of these html strings, which start differing at the stable id parameter; and so orders them according to stable ids.

**Solution**
Consists of two steps:
1) use the `data-order` attribute of the `td` element to define the data by that will be used for ordering table rows
2) since the value of the `data-order` attribute is a string, pad it with enough zeroes to ensure correct comparison (e.g. 2 becomes `00002` and is placed before `00010`, whereas otherwise string `2` would have been placed after string `10`)

## Views affected
`/:species/Transcript/Domains/Genes`

[Example in sandbox](http://ves-hx2-76.ebi.ac.uk:8410/Mus_musculus/Transcript/Domains/Genes?db=core;domain=IPR026271;g=ENSMUSG00000028591;r=4:100000001-150000000;t=ENSMUST00000030326)

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5292

### DIFF
--- a/modules/EnsEMBL/Web/Component/Transcript/DomainGenes.pm
+++ b/modules/EnsEMBL/Web/Component/Transcript/DomainGenes.pm
@@ -98,12 +98,15 @@ sub content {
     { key => 'desc', title => 'Description (if known)', width => '55%', align => 'left' }
   );
   
+  my $row_index = -1;
   foreach my $gene (sort { $object->seq_region_sort($a->seq_region_name, $b->seq_region_name) || $a->seq_region_start <=> $b->seq_region_start } @$genes) {
+    # the result of the sorting is the same as sorting by Genome Location column in the data table on the client; so we want to preserve this sorting order in a data attribute of the 'loc' column
+    $row_index += 1;
     my $row       = {};
     my $xref_id   = $gene->display_xref ? $gene->display_xref->display_id : '-';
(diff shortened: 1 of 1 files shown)


ens-st3:
OK

jyothishnt:
Could you please restart your sandbox as the change to add the chromosome:start-end is not visible atm. Thanks.

PR #731, opened 2019-09-23
Add sticky headers to DataTables

azangru:
## Description
1. Add sticky headers for data tables. Data tables are wrapped in all sorts of wrappers that prevent use of a simple pure CSS solution for sticky headers; so the headers were made sticky using javascript. When the table that is higher than the window height is scrolled past the top border of the screen, the top row of column headings will still be visible at the top of the screen to make it easier to navigate the columns of the table.
2. Add sticky headers for new tables. This was done purely with CSS (`position: sticky; top: 0`). I tested that on a gene variant table, where the sticky header works fine. Won't work in browsers that do not support this CSS rule (e.g. in IE11).

## Views affected
- Any views containing data tables.
E.g.: http://ves-hx2-76.ebi.ac.uk:8410/Homo_sapiens/Gene/Phenotype?db=core;g=ENSG00000139618;r=13:32315086-32400266

- Views containing new tables
E.g. BRCA2 variant table: http://ves-hx2-76.ebi.ac.uk:8410/Homo_sapiens/Gene/Variation_Gene/Table?db=core;g=ENSG00000139618;r=13:32315086-32400266

## Possible complications
Can't dare to imagine. Data tables can break in unexpected ways in unexpected places. From what I've tested, though, they look fine. Sticky headers in data tables (i.e. implemented with javascript rather than CSS) even work fine in IE11.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5052

### DIFF
--- a/htdocs/components/05_DataTable.js
+++ b/htdocs/components/05_DataTable.js
@@ -533,6 +533,8 @@ function StickyHeader (table) {
   this.tableHead = table.querySelector('thead');
   this.tableBody = table.querySelector('tbody');
 
+  this.scrollHandler = this.syncHeadScroll.bind(this);
+
   this.observeBody();
 }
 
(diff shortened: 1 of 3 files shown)


jyothishnt:
As discussed, here is the example to look at.
On RID view, Click
Configuration > Regulation > FeaturesByCellLines, Select All experiments and cells and go to Configuration.

azangru:
@jyothishnt I am afraid the approach using CSS position sticky won't work for data tables. Here's why.

Position: sticky seems to refer to the nearest scrollable parent. In the case of regulation matrix, the parent has a fixed height, so that you can scroll within it. In such scenario, position sticky works. Here's [a little demo](https://jsfiddle.net/kx80cr3u/) reproducing your setup.

If the nearest scrollable parent is the body itself, and there are no other parents with the `overflow` CSS rule, then `position: sticky` will also work fine. Here's a [demo](https://jsfiddle.net/kx80cr3u/1/) showcasing this.

However, if there is a parent with an `overflow` CSS rule and unset height, then `position: sticky` will just not work ([demo](https://jsfiddle.net/kx80cr3u/2/)). Annoyingly, it won't work even if we try to stick the element to the top or bottom of the page, and the `overflow` rule regulates only the horizontal scrolling of the parent ([demo](https://jsfiddle.net/kx80cr3u/3/)).

This last scenario is what we are dealing with in case of datatables. Their header is inside a container that has an overflow property, but does not have a height property; and therefore position: sticky doesn't work for these tables.

azangru:
@ens-st3 this link shows results of a VEP run in the sandbox for me:
http://ves-hx2-76.ebi.ac.uk:8410/Homo_sapiens/Tools/VEP/Results?tl=rAmaUoLPZ9aguyxz-1008

jyothishnt:
Hi @azangru, checked your sandbox for VEP results and variant page that uses the same datatable code. Looks good. Please go ahead and merge.

PR #732, opened 2019-09-23
Hide strains from other species in the orthologues tables

msmanoj:
## Description
Orthologues tables should only display the strains from the same strain set. (Eg. Pigs should not show Mouse strains)

## Views affected
http://ves-hx2-76.ebi.ac.uk:42228/Mus_spretus/Gene/Strain_Compara_Ortholog?collapse=none;db=core;g=MGP_SPRETEiJ_G0031873;r=8:44062-175441

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5275

### DIFF
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -90,11 +90,6 @@ sub content {
       delete $not_seen{lc $_};
     }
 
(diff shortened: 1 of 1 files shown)


ens-ap5:
Looks fine, though maybe the new code should have an explanatory comment for consistency?

msmanoj:
@ens-ap5 sorry, I have now removed the first `if` statement as it is not required. Would the comment I've added be sufficient?

ens-ap5:
Yes, I think so!

PR #735, opened 2019-10-07
Add CLS zmenu for long reads tracks

imran-ebi:
## Description

Long reads tracks have different zmenu criteria and did not have a custom zmenu for them. CLS is the specified zmenu in webdata for these tracks, though they actually didn't exist in the code. So a CLS zmenu module was created for this. There is also a minor fix for the long reads menu entry and removing of track code that was separating the long reads tracks (as they are now separated using web data).

## Views affected

Control panel and long reads zmenu

## Possible complications

Not sure

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5226


imran-ebi:
I've created separate CLS modules for gene and transcript. Also, removed havana_tangene check from the source label.

ens-ma7:
Looks good to me and agree having separate modules for gene and transcript.

PR #737, opened 2019-11-08
Removed extra word

muffato:
There is a spurious "either"


muffato:
I don't understand what the conflict is. I only have 1 commit and it's on top on master :thinking:

muffato:
Ah, GitHub now says there are no conflicts :man_shrugging:

PR #745, opened 2020-01-22
Fix ConfigMatrix incorrect numbers bug

msmanoj:
## Description
When configuring the RNASeq tracks, the numbers in the "tickboxes" are not correct. When not select it usually shows "0 1" and it shows "2 1" when selected. I would have expected "1 1" when the box is selected.

## Views affected
http://ves-hx2-76.ebi.ac.uk:42229/Homo_sapiens/Gene/Summary?g=ENSG00000139618;r=13:32315086-32400266

Configure this page -> RNASeq models

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5406


ens-ma7:
Waiting for sandbox to be up.

msmanoj:
Hi @ens-ma7, the sandbox is now up.

ens-ma7:
Looks like your fix broke the all enable/disable all link under default style

ens-ma7:
It turns out the "enable/disable all" link is a different problem which is also present on the live site. This  might require a different Jira. The fix for the current PR works fine and the counts are all good.

PR #746, opened 2020-01-23
Preserve info about original uri when dismissing request

azangru:
## Description
**Problem:** The 404 page is rendered when the apache handler dismisses the request.  At this point, apache re-routes the request to `/Error`, which is handled by the Apache::Error module (is it a good idea to re-route the request? probably not; but I'm not even attempting to touch this in this PR). As a result of this re-routing, the request object loses information about its initial uri; so the 404 page cannot behave differently depending on which page the user is coming from.

**Solution:** Add a custom header containing the original uri to the request object before dismissing it. This header can then be used to customise the 404 page depending on the requested url.

## Views affected
404 page

ENSWEB-5233

sandbox: http://ves-hx2-76.ebi.ac.uk:8430/

### DIFF
--- a/conf/httpd.conf
+++ b/conf/httpd.conf
@@ -454,7 +454,7 @@ ServerSignature         Off
 
 # Handle CGI errors with EnsEMBL::Web::Apache::ServerError
 ErrorDocument 500 /Crash
-# Handle missing pages with EnsEMBL::Web::Apache::NotFound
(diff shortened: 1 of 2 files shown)


ens-ap5:
Looks fine to me.

azangru:
Should this commit go to master or to postreleasefix/99?

PR #753, opened 2020-03-02
Add option to display announcement banner

msmanoj:
## Description
Added the option to display an announcement banner by adding a file to the tmp directory.

Steps to add the banner:
- Look for the ENSEMBL_TMP_MESSAGE_FILE in the SiteDefs and get the path.
- Create (if not exists) the file `ensembl_tmp_message` in the above path
- add an entry for `banner_message`.
- For example: `banner_message Hello World!` will display `Hello World!` inside the banner

## Views affected
Masthead

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5494

### DIFF
--- a/htdocs/components/99_announcement_banner.css
+++ b/htdocs/components/99_announcement_banner.css
@@ -1,0 +1,29 @@
+/*
+ * Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+ * Copyright [2016-2018] EMBL-European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#announcement-banner {
+    background-color: #CD0000;
+    padding: 12px 8px 12px 8px;
+    color: #FFF;
+    text-align: center;
+}
+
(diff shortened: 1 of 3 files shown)


modules/EnsEMBL/Web/Template/Legacy.pm
```diff
@@ -21,6 +21,7 @@ package EnsEMBL::Web::Template::Legacy;
 
 ### Legacy page template, used by standard HTML pages
 
+use EnsEMBL::Web::Utils::FileHandler qw(file_get_contents);
```
tuanebi: Should we remove this import?

htdocs/components/99_announcement_banner.css
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+
+#announcement-banner {
+    background-color: #CD0000;
+    padding: 12px 8px 12px 8px;
+    color: #FFF;
+    text-align: center;
+}
```
tuanebi: @ens-st3 I think Steve will decide the default colours for the banner on live site.

modules/EnsEMBL/Web/Template/Legacy.pm
ens-ap5: I think so - the TmpMessage module should be reading the file, not the template.

msmanoj: Removed the import

htdocs/components/99_announcement_banner.css
msmanoj: @tuanebi I sent the screenshot to Steve with the new colours we tried and he has Okayed it.

ens-st3: @msmanoj - can you confirm that we're not showing the pop up window ? That is to only be used in the few days before tools falls back to HH, everything else is to be communicated by the masthead banner

msmanoj: @ens-st3 Yes, I confirm that we are not displaying the pop up window. Also, the banner will start to appear only after creating the temporary file with the banner content, which we will be doing tomorrow.

ens-st3:
Ah that's new. Interesting

PR #758, opened 2020-03-10
Uopdate banner colours to match fallback

msmanoj:
Updated the announcement banner colour scheme to match the banner on the fallback site.

Current view:
http://uswest.ensembl.org/

Updated view:
http://ves-hx2-76.ebi.ac.uk:42228/index.html

### DIFF
--- a/htdocs/components/99_announcement_banner.css
+++ b/htdocs/components/99_announcement_banner.css
@@ -17,9 +17,9 @@
 
 
 #announcement-banner {
-    background-color: gold;
+    background-color: #dc0b0b;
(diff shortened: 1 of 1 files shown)


msmanoj:
@tuanebi do I need to make any changes to the public-plugins to update the banner for biomart?

tuanebi:
This colour is for limited functionality.
If we want to apply temporarily to the mirrors, I suggest to do it EC2_ebi plugin

msmanoj:
@tuanebi No, this is not going to be a temporary change just for mirrors. I guess we will be using the colours from limited functionality now and update it all later on.

PR #768, opened 2020-04-27
fix clinvar url - ENSVAR-3193 -e100

ima23:
## Description

_Using one or more sentences, describe the proposed changes and the reason for making them._

Variation summary section contains variation synonyms and link out to the original source. The ClinVar synonym link out is broken. This PR is to fix it.

## Views affected

Variation summary section and all views that contain it.

## Possible complications

None expected.

## Merge conflicts

None.

https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3193


jyothishnt:
Will merge and rebuild post-release

PR #769, opened 2020-05-04
Retreat to non-cache hub add if hub contains nonsense.

ens-ds23:
## Requirements

Hotfix for datahubs. Particularly high-priority because of forthcoming covid-site release.

Added multiple reviewers just prospectively. Feel free to look if you have the time, but don't feel obliged!

## Description

This fixes part of the problem with trackhub attachment for e100.

If bad data gets into the cache such that a cache get fails, then the get is retried without the cache. This limits the scope of cache get failures to the period when requests fail. The cache is populated in either case so if the request now succeeds the cache will be corrected.

A downside is that deleted datahubs will be contacted each time a page is loaded in-case they have sprung into life.

## Views affected

Region in detail (wiuth tracks attached)

## Possible complications

See Description section, last para.

## Merge conflicts

None envisaged.

None. Came through on slack from Jane.

### DIFF
--- a/modules/EnsEMBL/Web/Utils/TrackHub.pm
+++ b/modules/EnsEMBL/Web/Utils/TrackHub.pm
@@ -222,7 +222,7 @@ sub get_hub {
   my ($self, $args) = @_;
 
   my $out = $self->get_hub_internal($args,1);
-  if($out->{'error'}) {
+  if(!$out || $out->{'error'}) {
     $out = $self->get_hub_internal($args,0);
   }
   return $out;
--- a/modules/EnsEMBL/Web/Utils/TrackHub.pm
+++ b/modules/EnsEMBL/Web/Utils/TrackHub.pm
@@ -78,13 +78,18 @@ sub url {
   return $self->{'url'};
 }
 
-sub get_hub {
+sub get_hub_internal {
 ### Fetch metadata about the hub and (optionally) its tracks
 ### @param args Hashref (optional) 
 ###                     - parse_tracks Boolean
 ###                     - assembly_lookup Hashref
-### @return Hashref               
-  my ($self, $args) = @_;
+### @return Hashref              
+###
+### This method can be run with or without the ability to check the cache.
+### A wrapper method (called just get_hub) tries it first with and, if there
+### is an error, then without. This means that caches shouldn't get "poisoned"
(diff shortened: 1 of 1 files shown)


modules/EnsEMBL/Web/Utils/TrackHub.pm
```diff
@@ -204,6 +209,25 @@ sub get_hub {
...
+###                     - assembly_lookup Hashref
+### @return Hashref              
+###
+### This method calls get_hub_internal first with and, if there
+### is an error, then without. This means that caches shouldn't get "poisoned"
+### if they fail due to an intermittent failure on first population. 
+  my ($self, $args) = @_;
+
+  my $out = $self->get_hub_internal($args,1);
+  if($out->{'error'}) {
```
tuanebi, line 225: What happen if $out is empty or broken somehow without 'error' field?

tuanebi:
In the original get_hub method, I understand `$trackhub` is only cached when there is no error detected, from the block:
https://github.com/Ensembl/ensembl-webcode/blob/release/100/modules/EnsEMBL/Web/Utils/TrackHub.pm#L194-L203

But the error on live website suggests `$cache` also keeps a bad copy of `$trackhub`. If this is true, does it mean the logic in that block needs to be updated?

modules/EnsEMBL/Web/Utils/TrackHub.pm
ens-ds23: Good point. I'll make this test stronger.

ens-ds23:
It's strange, isn't it? I think what must happen is that it seems correct and with no error, gets written to memcached but is somehow corrupted there but the key is set to something (bad) or is set to a good value that somehow can't be read.

I think it's ok to have both, just to be sure. But there's a lot about this bug I still don't understand.

ens-ds23:
Made @tuanebi 's suggested fix. My testing is, on a sandbox:
1. make sure attaching trackhubs works with new fix.
2. deliberately break cache response by adding nonsense for variable after cache retrieval and verify still works;
3. effectively disable the fix in this PR (by short-circuiting the "if") with that deliberate break still in-place and verify things would break.

When that's done, I'll push live.

ens-ds23:
Done the tests above. About to push live.

PR #770, opened 2020-05-12
Hide gene tree when NO_COMPARA

msmanoj:
## Description

Hide gene tree when NO_COMPARA is set to 1.

## Views affected
http://ves-hx2-76.ebi.ac.uk:42227/Sars_cov_2/Info/Index?db=core;g=ENSSASG00005000003;r=MN908947.3:266-13483;t=ENSSAST00005000003
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5709

### DIFF
--- a/modules/EnsEMBL/Web/Component/Info/HomePage.pm
+++ b/modules/EnsEMBL/Web/Component/Info/HomePage.pm
@@ -251,11 +251,17 @@ sub genebuild_text {
 
 sub compara_text {
   my $self         = shift;
+
+  if($SiteDefs::NO_COMPARA eq 1){
+    return '';
+  }
+  
   my $hub          = $self->hub;
   my $species_defs = $hub->species_defs;
   my $sample_data  = $species_defs->SAMPLE_DATA;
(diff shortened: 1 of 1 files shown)


ens-st3:
I thought this should hide automatically. Is there an example gene tree in the database or .INI file or does it reuse the example gene ?

ens-ap5:
> I thought this should hide automatically.

No, we assume that if there's a compara database (which there always is on the current sites), there will be a gene tree example. Maybe we should check for an example in the SAMPLE_DATA hash instead, as that would be more robust going forward (especially with a trimmed-down compara).

msmanoj:
> Please remove "eq 1" from the if condition. In Perl, "eq" is only used for strings, not numbers, and in any case the value of the test "if ($SiteDefs::NO_COMPARA)" will be interpreted as a boolean.

Thanks @ens-ap5. That's now done.

jyothishnt:
should we override this in covid19 plugin @ens-ap5 ? Or leave it in the feature branch?

PR #771, opened 2020-05-18
Fix population graphs

azangru:
### Steps to reproduce
Note the bug depends on the state of the browser (probably cookies, although I have not investigated this further)
-  Visit http://www.ensembl.org/Homo_sapiens/Variation/Population?db=core;r=1:230709548-230710548;v=rs699;vdb=variation;vf=179
- Clear site data for ensembl.org (in Chrome: open dev tools > Application > Clear site data)
- Refresh the page (make sure that there is no hash value in the end of the url)
- You should see population charts:
image 
- Open one of the sub-populations
- Refresh the page again
- The charts will likely stop rendering
image

### Cause of the bug
If the browser does not have certain values preserved in its state (🤷‍♂️) then loading the page will trigger the `toggleable` function in Content.js, which will [simulate click](https://github.com/Ensembl/ensembl-webcode/blob/release/100/htdocs/components/10_Content.js#L268) event on any elements that have been "opened", such as the subpopulation on this page. This will fire the [toggleContent](https://github.com/Ensembl/ensembl-webcode/blob/release/100/htdocs/components/15_PopulationGraph.js#L33-L43) method on PopulationGraph, which will call the [makeGraphs](https://github.com/Ensembl/ensembl-webcode/blob/release/100/htdocs/components/15_PopulationGraph.js#L36) function. The problem here is that for the `makeGraphs` function to work, the [getContent](https://github.com/Ensembl/ensembl-webcode/blob/release/100/htdocs/components/10_Piechart.js#L33-L59) method must first have executed; otherwise there won't be necessary data in the Piechart class, and `makeGraphs` will fail.

### Solution
Check whether the `getContent` method has completed (by an attribute that it sets on the Piechart class) before running `makeGraphs`. It may not be a very pretty solution, but since this code will eventually be replaced, it's probably ok.

### Another change
Remove the hash anchor from the url when a toggleable element is turned off.

## Views affected
Population genetics for variant.

View in sandbox: http://ves-hx2-76.ebi.ac.uk:8410/Homo_sapiens/Variation/Population?db=core;r=1:230709548-230710548;v=rs699;vdb=variation;vf=179

## Possible complications
🤷‍♂️ 

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5503

### DIFF
--- a/htdocs/components/10_Content.js
+++ b/htdocs/components/10_Content.js
@@ -244,6 +244,9 @@ Ensembl.Panel.Content = Ensembl.Panel.extend({
           if (panel.elLk[this.rel][0].id) {
             window.location.hash = panel.elLk[this.rel][0].id;
           }
+        } else {
+          // remove the hash from the url
+          history.replaceState("", document.title, window.location.pathname + window.location.search);
         }
       }
       
--- a/htdocs/components/15_PopulationGraph.js
+++ b/htdocs/components/15_PopulationGraph.js
@@ -29,10 +29,13 @@ Ensembl.Panel.PopulationGraph = Ensembl.Panel.Piechart.extend({
     
     this.base();
   },
-  
+
   toggleContent: function (el) {
     if (el.hasClass('closed') && !el.data('done')) {
       this.base(el);
+      if (!this.canAddContent()) {
+        return;
+      }
       this.makeGraphs(this.el.find('.' + el.attr('rel') + ' .pie_chart > div[id^=graphHolder]').map(function() { return this.id.match(/\d+/).pop() }).toArray());
       el.data('done', true);
     } else {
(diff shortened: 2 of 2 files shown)


ens-ma7:
@azangru 
By removing the hash from the URL that means when you refresh the page you wont be able to go that section again? 

I am not sure i understand your solution. what about using the defer function to execute the makegraph after getcontent has been done?

azangru:
> By removing the hash from the URL that means when you refresh the page you wont be able to go that section again?

Yes. But the hash should get removed when a toggleable element is "closed". That should signal the user's intent not to view this section anymore.

> I am not sure i understand your solution. what about using the defer function to execute the makegraph after getcontent has been done?

I am preventing the PopulationGraph module from executing `toggleContent` on initial page load entirely. It seems that there is no need for it to do that anyway. The way this works is that PopulationGraph checks for the field (`this.graphEls`) that gets created during execution of `getContent`. If there is no `this.graphEls` field, `this.makeGraphs` should not be executed.

PR #775, opened 2020-06-27
fix species selector bug due to recent changes in SpeciesDefs

jyothishnt:
## Description

Species selector not adding species to its proper clade.
Reason is that it SPECIES_IMAGE_DIR is not set. This was added for RR site and is undef for the current site there by not setting the SPECIES_IMAGE variable.

## Views affected

Species selector in BLAST, Alignments and AlignSlice

## Possible complications

Need to check why $config_packer->tree->{ASEEMBLY_VERSION} is not available here https://github.com/Ensembl/ensembl-webcode/compare/hotfix-sp-selector?expand=1#diff-93e79820da11cca1db9ee8ce3e7b1811L899

We will have to cherry pick this to master branch because I couldn't do it in postreleasefix/100 branch as this RR SPECIES_IMAGE addition commit is only available in master and release/100. This created a fix for release/100

## Merge conflicts

No

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5873

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -200,6 +200,7 @@ our @ENSEMBL_CONF_DIRS               = ("$ENSEMBL_WEBROOT/conf");               
 our @ENSEMBL_HTDOCS_DIRS             = ($ENSEMBL_DOCROOT, "$ENSEMBL_SERVERROOT/biomart-perl/htdocs");  # locates static content
 our $ENSEMBL_TAXONOMY_DIVISION_FILE  = ("$ENSEMBL_DOCROOT/e_divisions.json");
 our $DEFAULT_SPECIES_IMG_DIR         = 'htdocs/i/species';
+our $SPECIES_IMAGE_DIR               = defer { $SiteDefs::ENSEMBL_SERVERROOT.'/public-plugins/ensembl/'. $SiteDefs::DEFAULT_SPECIES_IMG_DIR };
 ###############################################################################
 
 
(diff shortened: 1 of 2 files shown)


conf/SiteDefs.pm
```diff
@@ -200,6 +200,7 @@ our @ENSEMBL_CONF_DIRS               = ("$ENSEMBL_WEBROOT/conf");
 our @ENSEMBL_HTDOCS_DIRS             = ($ENSEMBL_DOCROOT, "$ENSEMBL_SERVERROOT/biomart-perl/htdocs");  # locates static content
 our $ENSEMBL_TAXONOMY_DIVISION_FILE  = ("$ENSEMBL_DOCROOT/e_divisions.json");
 our $DEFAULT_SPECIES_IMG_DIR         = 'htdocs/i/species';
+our $SPECIES_IMAGE_DIR               = defer { $SiteDefs::ENSEMBL_SERVERROOT.'/public-plugins/ensembl/'. $SiteDefs::DEFAULT_SPECIES_IMG_DIR };
```
jyothishnt, line 203: I think we dont need to do this. I have seen a PR in public_plugins from Manoj which got approved. So I will update this PR. But please do check the remaining ones.

ens-ap5: Yes, this needs to go into public-plugins, since many sites won't have this file location available.

jyothishnt:
Thats done

PR #786, opened 2020-10-02
Fix TSL popup issue on transcript table.

mehrnazcharkhchi:
## Requirements

## Description
@ens-ma7 's explained the cause of this issue in the related Jira ticket:
_Since we moved the help system to using the EBI glossary (https://www.ebi.ac.uk/ols/api/ontologies/ensemblglossary/terms?size=500 ), the key for retrieving the help content for TSL has changed. Instead of TSL:1 it is now "TSL 1" and for TSL it is "Transcript support level"._

## Views affected

Transcript table TSL popup:
http://www.ensembl.org/Homo_sapiens/Gene/Summary?g=ENSG00000139618;r=13:32315086-32400268

## Possible complications
_

## Merge conflicts
_

_ Sandbox link is added to the following ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5684

### DIFF
--- a/modules/EnsEMBL/Web/Component/Shared.pm
+++ b/modules/EnsEMBL/Web/Component/Shared.pm
@@ -308,8 +308,8 @@ sub transcript_table {
         }
         if ($trans_attribs->{$tsi}{'TSL'}) {
           my $tsl = uc($trans_attribs->{$tsi}{'TSL'} =~ s/^tsl([^\s]+).*$/$1/gr);
-          push @flags, helptip("TSL:$tsl", get_glossary_entry($hub, "TSL:$tsl").get_glossary_entry($hub, 'TSL'));
-        }
+          push @flags, helptip("TSL:$tsl", get_glossary_entry($hub, "TSL $tsl").get_glossary_entry($hub, 'Transcript support level'));
+	}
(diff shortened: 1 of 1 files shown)


ens-ma7:
Hi @Mehrnaz-Charkhchi code change is fine but you made the PR against the branch postreleasefix/101, is that correct? We only made a PR against a postreleasefix branch if we want to push live.  Looking at the Jira ticket, it says for release 102.

The other thing that i will advise is to add some description in the PR and the Jira ticket about the cause of the problem and what was the fix. I have added a description to the Jira ticket, might be good to add something in the PR as well for future reference.

mehrnazcharkhchi:
Thanks @ens-ma7, You're right it's a 102 fix. 
Should I wait for 102 branches?
Thanks for adding description to the ticket. I'll add details to PR as well.

ens-ma7:
@Mehrnaz-Charkhchi you can do the PR against master.  It will get push to the release branch when we create it later in the release cycle.

mehrnazcharkhchi:
@ens-ma7, Done!

ens-ma7 (approved):

PR #787, opened 2020-10-15
Fix cactus alignment - out of memory error

msmanoj:
## Description

This PR is to fix the error displayed in the following page:
http://www.ensembl.org/Mus_musculus/Location/Compara_Alignments/Image?align=835;db=core;r=4%3A136366473-136547301;time=1596020339619.619

I have increased the PERL_RLIMIT_AS value from `2048` to `2560` to fix this issue.

## Views affected
Temporarily fixed this on staging:
http://staging.ensembl.org/Mus_musculus/Location/Compara_Alignments/Image?align=835;db=core;r=4%3A136366473-136547301;time=1596020339619.619

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5908

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -95,7 +95,7 @@ our $ENSEMBL_CONFIG_BUILD             = 0; # Build config on server startup? Set
 our $ENSEMBL_SERVER_SIGNATURE         = "$ENSEMBL_SERVER-$ENSEMBL_SERVERROOT" =~ s/\W+/-/gr; # Unique string representing this machine/server
 our $ENSEMBL_SITETYPE                 = 'Ensembl';
 our $ENSEMBL_HELPDESK_EMAIL           = defer { $ENSEMBL_SERVERADMIN };   # Email address for contact form and help pages
-our $PERL_RLIMIT_AS                   = '2048:4096';                      # linux does not honor RLIMIT_DATA, RLIMIT_AS (address space) will work to limit the size of a process
+our $PERL_RLIMIT_AS                   = '2560:4096';                      # linux does not honor RLIMIT_DATA, RLIMIT_AS (address space) will work to limit the size of a process
 our $ENSEMBL_REST_URL                 = 'http://rest.mydomain.org';       # url to your REST service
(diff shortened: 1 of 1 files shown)


ens-st3 (approved):
Are there any potential downsides to doing this @ens-ds23 ?

ens-ds23:
I need to check this to see what's happening in terms of the fix as it's counter-intuitive to me as this changes the soft limit (first number) which should never prevent any requests completing successfully. So this fix seems to work in some odd way: so I'll read the ticket, test it out, etc.

The potential issue is that it impacts (in a complex way) on the sharing of resources on the machine which could make things more crashy (or less!) under load. So we should really understand the impact first.

My intuition is that the change will prove ok, though.

ens-ds23:
Well, it definitely works! But how? ...

msmanoj:
> Well, it definitely works! But how? ...

@ens-ds23 - here is the similar ticket thats about the same error:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-3959

ens-ds23:
It looks like soft/hard here (via Apache2::Resource) is being used in the setrlimit(2) sense unlike in the rest of the code and the rest code. So we can ignore the hard limit (as long as it's some value more than soft) and understand that the "soft" limit is the actual limit. So this fix makes sense. I'll just have a think on if it has any impact on the live machines in terms of memory.

ens-ds23:
There's not an awful lot of overhead on the www machines and it looks like for some reason we don't OOB limit memory on www at all (as we do on grch37, mirrors, etc). Setting an OOB limit to a lower value would allow the worker to finish this request but then immediately quit. This would mean that very big workers don't hang around for ages after someone's been busy with cactus stuff. Enough of that could crash these machines otherwise. I think this is probably true at the moment, too, but less so. I'll work out a patch that can be added to his PR.

ens-ds23:
This PR is good, I think. I will put in another into ebi-plugins to set an OOB memory limit for www. We should make sure they both go in though (or neither).

ens-ds23 (approved):
approved. see comments.

ens-ds23:
companion bugfix in ebi-plugins is #65 https://github.com/Ensembl/ebi-plugins/pull/65

PR #788, opened 2020-10-21
Update transcript table to display uniprot isoform

msmanoj:
## Description

- Renamed the column name to `UniProt Match`
- `UniProt Match` column will no display the informs if they exist and shows the gene level links if not.
- Added external link entry to DEFAULTS.ini to `UNIPROT_ISOFORM`

## Views affected

http://ves-hx2-75.ebi.ac.uk:42228/Homo_sapiens/Gene/Summary?g=ENSG00000159840;r=7:143381295-143391111

http://ves-hx2-75.ebi.ac.uk:42228/Homo_sapiens/Gene/Matches?g=ENSG00000159840;r=7:143381295-143391111

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5877

### DIFF
--- a/conf/ini-files/DEFAULTS.ini
+++ b/conf/ini-files/DEFAULTS.ini
@@ -574,7 +574,6 @@ UNIPROT                  = http://www.uniprot.org/uniprot/###ID###
 UNIPROTKB/SWISS-PROT        = http://www.uniprot.org/uniprot/###ID###
 UNIPROTKB/TREMBL            = http://www.uniprot.org/uniprot/###ID###
 UNIPROT/VARSPLIC            = http://www.uniprot.org/uniprot/###ID###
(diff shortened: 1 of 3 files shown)


ens-ap5 (changes requested):
Mostly looks good, but I don't think you need the URL template twice in DEFAULTS.ini - that's one big long list that needs tidying up! (SwissProt and TREMBL shouldn't be in there twice either). If you scroll to line 428 you'll see a heading 'RETIRE?' - when I reorganised the list several years ago, I moved everything I knew was in use out of this section, but didn't spot the duplicate values.

ens-st3:
Table looks good. Let's wait until OLS is updated to confirm how the gene summary line should be worded and what data it should show

ens-ap5 (approved):

ens-st3 (approved):

PR #794, opened 2020-11-10
Update eQTL displays to use new eQTL catalog REST endpoints rather than e! REST

msmanoj:
## Description
We are decommissioning the eQTL endpoints available under Ensembl REST and start to use the GTEX data added to eQTL catalog REST endpoint.

**eQTL catalog REST endpoint docs:**
http://www.ebi.ac.uk/eqtl/api-docs/

**Here is a document with details about the current and the new endpoint:**
https://docs.google.com/document/d/1FbxgKPGjQNdiP7l04btwDpC8M-u5sXLMYj96dmGn1B0/edit#heading=h.qqrby2wpp9yw

**NOTE**
The following issue is still pending to be resolved:
https://github.com/eQTL-Catalogue/eQTL-SumStats/issues/22
I have checked this with Daniel Zerbino and James Hayhurst and they have confirmed that we are okay to proceed with this for now.
Without the `display_consequence` field, the plots will all be greyed out as shown below:

Screenshot 2020-11-10 at 14 46 47

Currently on live:
Screenshot 2020-11-10 at 14 47 16

## Views affected

**Regulation Manhattan Plot**
http://ves-hx2-75.ebi.ac.uk:42229/Homo_sapiens/Gene/Regulation?g=ENSG00000139618;r=13:32315086-32400268

**Gene expression correlations table**
http://ves-hx2-75.ebi.ac.uk:42229/Homo_sapiens/Variation/Mappings?r=1:230709548-230710548;v=rs699;vdb=variation;vf=179

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5483
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5282
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5893
https://www.ebi.ac.uk/panda/jira/browse/ENSINT-415
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5399

### DIFF
--- a/modules/EnsEMBL/Web/ImageConfig/generegview.pm
+++ b/modules/EnsEMBL/Web/ImageConfig/generegview.pm
@@ -29,6 +29,7 @@ use List::MoreUtils qw(any);
 sub init_cacheable {
   ## @override
   my $self = shift;
+  my $hub  = $self->hub;
 
   $self->SUPER::init_cacheable(@_);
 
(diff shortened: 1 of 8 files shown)


ens-ap5 (changes requested):
For consistency, the new REST URL should go into the same plugin as the main REST URL - note that the core SiteDefs contains a placeholder URL for REST. I think that it should go into public-plugins/ensembl, but you'd need to check.

msmanoj:
> For consistency, the new REST URL should go into the same plugin as the main REST URL - note that the core SiteDefs contains a placeholder URL for REST. I think that it should go into public-plugins/ensembl, but you'd need to check.

@ens-ap5  - Unlike REST URL, the eQTL REST url will always be the same across all sites so I'm wondering if it is worth having just a placeholder in the core SiteDefs and overriding in all other places.
I also noticed a similar entry where we have not used a placeholder:
https://github.com/Ensembl/ensembl-webcode/blob/2686b6197ed670e4b5aa29ba48719ff872b34070/conf/SiteDefs.pm#L233

ens-ap5:
Do we even have eQTL data outside www? If not, I see no point in having it in the main SiteDefs and then overriding it elsewhere.

msmanoj:
> Do we even have eQTL data outside www? If not, I see no point in having it in the main SiteDefs and then overriding it elsewhere.

No, we don't. Do we need to define it somewhere else so that it is specific to www?

ens-st3:
We have it for grch37, archives, and externally maintained mirrors

modules/EnsEMBL/Web/ImageConfig/generegview.pm
```diff
@@ -56,6 +56,8 @@ sub init_cacheable {
     $gtex_tissue_example = $gtex_tissues[0];
   }
 
+  warn "gtex_tissue_example: ". $gtex_tissue_example;
```
ens-st3: remove warn

ens-st3:
Gene Regulation page - can you remove 'ARRAY(0x37766098) GTEX eQTLs' from the list of available tracks.
Gene expression correlations - I'm afraid all I can see is 'undefined'
Screenshot 2020-11-23 at 18 40 47
Screenshot 2020-11-23 at 18 40 28

msmanoj:
> Gene Regulation page - can you remove 'ARRAY(0x37766098) GTEX eQTLs' from the list of available tracks.
@ens-st3  - That's done.
> Gene expression correlations - I'm afraid all I can see is 'undefined'
- I think this is just a caching issue. Could you please try again by clearing your cache?

PR #801, opened 2020-12-08
Replace deprecated method with get_ref_length

nwillhoft:
## Requirements

## Description

During sandbox web server testing, I came across a call to the deprecated (and removed) method `get_total_length`, which is currently in use at `ensembl-webcode/modules/EnsEMBL/Web/Component/Shared.pm`. This PR replaces the call to this method with the recommended replacement method, `get_ref_length`.

I came across this when my test web server consistently failed to load the Whole Genome Display web page for each species, and the error pointed towards this deprecated method. This web page now appears to load as normal with this fix in place.  

This PR is being submitted against `postreleasefix/102`, as discussed with @ens-ap5.

## Views affected

The Whole Genome Display page loads as normal with this fix. Please see JIRA ticket below for error example.

## Possible complications

None I'm aware of.

## Merge conflicts

N/A

https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3621

### DIFF
--- a/modules/EnsEMBL/Web/Component/Shared.pm
+++ b/modules/EnsEMBL/Web/Component/Shared.pm
@@ -761,7 +761,7 @@ sub species_stats {
   });
   $summary->add_row({
       'name' => '<b>Base Pairs</b>',
-      'stat' => $self->thousandify($genome_container->get_total_length()),
+      'stat' => $self->thousandify($genome_container->get_ref_length()),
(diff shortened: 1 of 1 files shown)


msmanoj (approved):

msmanoj:
Hi @mehrnazcharkhchi, can you get this PR merged, please?

PR #791, opened 2020-11-03
[ON HOLD] Fix strain orthologues downloads

msmanoj:
## Description

- Fixed the issue with breed orthologues downloads where it was working the same way as for the reference species.

The issue was with the following line:
https://github.com/Ensembl/ensembl-webcode/pull/791/files#diff-39a1865bcf5a112d5bb6d3e05f7c9c4fb0fd7791324c77ad8d36cdd63cf1a0c5R85
where it was adding all the available species to $hub->params

## Views affected

http://ves-hx2-75.ebi.ac.uk:42229/Sus_scrofa/Gene/Strain_Compara_Ortholog?db=core;g=ENSSSCG00000004244;r=1:42729112-42896283

http://ves-hx2-75.ebi.ac.uk:42229/Sus_scrofa/Gene/Compara_Ortholog?db=core;g=ENSSSCG00000004244;r=1:42729112-42896283

## Related JIRA Issues (EBI developers only)

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5895
https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4069

PR #802, opened 2020-12-10
Fix strain orthologues downloads

muffato:
## Description

(Supersedes #791)

Downloading data (with "Download orthologues" button) from the Strains' Orthologues view currently returns the orthologues of the main species set, not of the strains. This PR is to fix that and return the orthologues of the strains instead.

- The definition of `my $strain_tree` is inspired from https://github.com/Ensembl/ensembl-webcode/blob/release/102/modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm#L41 . It seems to work with just the test of `$hub->param('data_action')` and I don't know if the rest (`|| $self->hub->species_defs->IS_STRAIN_OF || $self->hub->param('strain')`) would be needed too.
- I have made the function call the existing `get_compara_Member` and `get_GeneTree` to avoid duplicating code / functionality. It looks like these two functions cache the data to speed up further calls. I imagine it's not a bad thing to have in the export view, but I'd like to double-check.

## Views affected

- Orthologues table for the Strains
  - http://ves-hx2-76.ebi.ac.uk:5080/Mus_musculus/Gene/Strain_Compara_Ortholog?g=ENSMUSG00000017167;r=11:101170523-101190724 is fixed
  - http://ves-hx2-76.ebi.ac.uk:5080/Mus_musculus/Gene/Compara_Ortholog?g=ENSMUSG00000017167;r=11:101170523-101190724 still works

## Possible complications

None expected.

## Merge conflicts

None expected.

- https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5895
- https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4069

### DIFF
--- a/modules/EnsEMBL/Web/Object/Gene.pm
+++ b/modules/EnsEMBL/Web/Object/Gene.pm
@@ -888,11 +888,12 @@ sub get_homologue_alignments {
   my $type        = shift || 'ENSEMBL_ORTHOLOGUES';
   my $database    = $self->database($compara_db);
   my $hub         = $self->hub;
+  my $strain_tree = $hub->species_defs->get_config($hub->species,'RELATED_TAXON') if($hub->param('data_action') =~ /strain_/i);
   my $msa;
 
   if ($database) {  
     my $member  = $self->get_compara_Member($compara_db);
-    my $tree    = $self->get_GeneTree($compara_db, 1);
+    my $tree    = $self->get_GeneTree($compara_db, 1, $strain_tree);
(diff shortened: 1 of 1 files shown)


ens-st3 (approved):
Checked a few downloads, looks good

ens-ap5 (approved):
Looks good. We need to review the use of RELATED_TAXON, as we should probably get the required parameter from the databases instead of the ini files, but that's a project for another day!

PR #808, opened 2021-01-21
Bug fix/ensweb 6115

ens-ds23:
## Description

Some alignsliceslices could not be serialized/deserialized successfully in some circumstances owing to representing truncated regions not directly present in the alignsliceslices within the alignslice itself. There is no good way to serialise alignsliceslices, this is something that was confirmed with compara at first implementation. An alignsliceslice resembles a slice (from the core API) but is not a fully featured slice. However it can be used in its place in most circumstances, supported by perl's loose typesystem. To this end, all the properties of the alignsliceslice _apart from_ the start and end co-ordinates are stored in a similarly fake coordsystem (referenced from the slice), species into species, etc, but those without a natural home go into "version". If we can't find an exact match (which we always try first) we get an alignsliceslice which matches in the coordsystem (by matching species and version) and then update its start and end coordinates.

## Views affected

http://ves-hx-78.ebi.ac.uk:8000/Cyprinus_carpio/Location/Compara_Alignments/Image?align=1801--Oryzias_latipes--1:13183763-13215284;db=core;g=ENSCCRG00000006701;r=LN590858.1:385743-420560

## Possible complications

Compara views slices might be confused in images (ie the wrong blocks aligned) if there's a bug in the code to compare coord-systems. We should check carefully before merging. This has been done somewhat (double-checking code and browser) and is continuing. It is less likely to affect views which actually worked prior to the bug being revealed because there is still ana ttempt to get an exact match first. However, the algorithm for matching has changed so this also needs to be checked.

Precache would benefit from rebuilding as these changes will make old precaches less effective.

## Merge conflicts

Currently against release/102, this PR will almost certainly be closed and moved when the right base branch is found!

ENSWEB-6115 RT#467081

### DIFF
--- a/modules/EnsEMBL/Web/Command/UserData/TrackHubRedirect.pm
+++ b/modules/EnsEMBL/Web/Command/UserData/TrackHubRedirect.pm
@@ -35,9 +35,6 @@ sub process {
   my $url           = $hub->param('url');
      $url           =~ s/(^\s+|\s+$)//g; # Trim leading and trailing whitespace
   my $filename      = [split '/', $url]->[-1];
-  my @bits          = split /\./, $filename;
-  my $extension     = $bits[-1] eq 'gz' ? $bits[-2] : $bits[-1];
-  my $pattern       = "^$extension\$";
   my ($redirect, $anchor);
   my $params        = {};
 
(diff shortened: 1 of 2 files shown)


ens-ma7 (approved):

ens-ma7:
@ens-ds23 is this PR still valid? I had a look and the code changes looks good and does what it expect even though its not the prettiest way of doing it.

ens-ds23:
I think we should move this onto master. My worries about breaking things would be lessened by putting this into a new release for the full QA.

ens-ds23:
I'll try to do the git magic now.

ens-ds23:
Please ignore the extra commit from Anne in this PR. I messed up the rebase. It should merge without hassle.

PR #809, opened 2021-01-26
Feature/collections

ens-ap5:
## Requirements

## Description

Add ability to handle collection databases, since we will be having these in the next release of the covid-19 website. Code has been manually incorporated from equivalent modules in eg-web-common.

## Views affected

Potentially any species-specific views, but as a first test, the main species table should contain the correct number of species: test.ensembl.org/info/about/species.html

## Possible complications

Serious - if databases are not parsed correctly, entire species or tracks could be missing.

## Merge conflicts

No conflicts at the moment.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5963

### DIFF
--- a/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -1661,8 +1661,8 @@ sub _munge_meta {
   while (my ($species_id, $meta_hash) = each (%$meta_info)) {
     next unless $species_id && $meta_hash && ref($meta_hash) eq 'HASH';
     
-    my $species  = $meta_hash->{'species.url'}[0];
-    my $bio_name = $meta_hash->{'species.scientific_name'}[0];
+    my $species         = $meta_hash->{'species.url'}[0];
+    my $production_name = $meta_hash->{'species.production_name'}[0];
     
     ## Put other meta info into variables
     while (my ($meta_key, $key) = each (%keys)) {
@@ -1676,11 +1676,11 @@ sub _munge_meta {
                   ? $self->db_tree->{'ASSEMBLY_VERSION'} : $value;
       }
 
-      $self->tree->{$key} = $value;
+      $self->tree($production_name)->{$key} = $value;
     }
 
     ## Uppercase first part of common name, for consistency
-    $self->tree->{'SPECIES_COMMON_NAME'} = ucfirst($self->tree->{'SPECIES_COMMON_NAME'});
+    $self->tree($production_name)->{'SPECIES_COMMON_NAME'} = ucfirst($self->tree->{'SPECIES_COMMON_NAME'});
 
     ## Do species group
     my $taxonomy = $meta_hash->{'species.classification'};
@@ -1688,9 +1688,9 @@ sub _munge_meta {
     if ($taxonomy && scalar(@$taxonomy)) {
       my %valid_taxa = map {$_ => 1} @{ $self->tree->{'TAXON_ORDER'} };
       my @matched_groups = grep {$valid_taxa{$_}} @$taxonomy;
-      $self->tree->{'TAXONOMY'} = $taxonomy;
-      $self->tree->{'SPECIES_GROUP'} = $matched_groups[0] if @matched_groups;
-      $self->tree->{'SPECIES_GROUP_HIERARCHY'} = \@matched_groups;
+      $self->tree($production_name)->{'TAXONOMY'} = $taxonomy;
+      $self->tree($production_name)->{'SPECIES_GROUP'} = $matched_groups[0] if @matched_groups;
+      $self->tree($production_name)->{'SPECIES_GROUP_HIERARCHY'} = \@matched_groups;
     }
 
     ## create lookup hash for species aliases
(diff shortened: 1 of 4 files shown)


ens-ap5:
Equivalent NV modules from which code has been copied:

https://github.com/EnsemblGenomes/eg-web-common/blob/release/eg/49/modules/EnsEMBL/Web/ConfigPacker.pm#L394
https://github.com/EnsemblGenomes/eg-web-common/blob/release/eg/49/modules/EnsEMBL/Web/SpeciesDefs.pm#L351

modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -1702,40 +1702,37 @@ sub _munge_meta {
...
     ## Used mainly in <head> links
-    ($self->tree->{'SPECIES_BIO_SHORT'} = $self->tree->{'SPECIES_URL'}) =~ s/^([A-Z])[a-z]+_([a-z]+)$/$1.$2/;
+    ($self->tree($production_name)->{'SPECIES_BIO_SHORT'} = $self->tree->{'SPECIES_URL'}) =~ s/^([A-Z])[a-z]+_([a-z]+)$/$1.$2/;
     
-    if ($self->tree->{'ENSEMBL_SPECIES'}) {
-      push @{$self->tree->{'DB_SPECIES'}}, $species;
-    } else {
-      $self->tree->{'DB_SPECIES'} = [ $species ];
-    }
+    push @{$self->tree->{'DB_SPECIES'}}, $species;
```
ens-ma7, line 1707: Just checking if we still need the if statement from the left.

modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -1934,5 +1937,10 @@ sub _munge_species_url_map {
   $multi_tree->{'ENSEMBL_SPECIES_URL_MAP'} = \%species_map;
 }
 
+sub is_collection {
+  my ($self, $db_name) = @_;
+  my $database_name = $self->tree->{'databases'}->{'DATABASE_CORE'}{'NAME'};
```
ens-ma7, line 1942: use $db_name instead of hardcoding DATABASE_CORE

modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -1809,10 +1806,16 @@ sub _munge_meta {
...
+    $self->tree($production_name)->{'SAMPLE_DATA'} = $sample_hash if scalar keys %$sample_hash;
 
     # check if the karyotype/list of toplevel regions ( normally chroosomes) is defined in meta table
-    @{$self->tree->{'TOPLEVEL_REGIONS'}} = @{$meta_hash->{'regions.toplevel'}} if $meta_hash->{'regions.toplevel'};
+    @{$self->tree($production_name)->{'TOPLEVEL_REGIONS'}} = @{$meta_hash->{'regions.toplevel'}} if $meta_hash->{'regions.toplevel'};
+
+    ## need to explicitly define as empty array by default
+    ## SpeciesDefs looks for a value at collection level
+    if ($self->is_collection('DATABASE_CORE')) {
+      @{$self->tree($production_name)->{'ENSEMBL_CHROMOSOMES'}} = ();                                     @{$self->tree($production_name)->{'ENSEMBL_CHROMOSOMES'}} = @{$meta_hash->{'region.toplevel'}} if $meta_hash->{'region.toplevel'};
```
ens-ma7, line 1817: Not a big thing but the styling is bothering me, why the @{$self...} after the semi colon going on the same line, it should be another line.

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -656,7 +656,8 @@ sub _expand_database_templates {
         $tree->{'databases'}{'DATABASE_'.$_} = $db_name;
       }
       else {
-        unless ($filename eq 'MULTI' && $SiteDefs::NO_COMPARA) {
+        ## Ignore this step for MULTI, as it may not have a core db
+        unless ($filename eq 'MULTI') {
```
jyothishnt, line 660: We dont use NO_COMPARA anymore?

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -878,7 +888,12 @@ sub _parse {
   my $datasets = [];
   my $aliases  = $tree->{'MULTI'}{'ENSEMBL_SPECIES_URL_MAP'};
   my $labels    = $tree->{'MULTI'}{'TAXON_LABEL'};  
-  foreach my $prodname (@$SiteDefs::PRODUCTION_NAMES) {
+
+  ## Loop through all keys, not just PRODUCTION_NAMES (need for collection dbs)
+  foreach my $key (sort keys %$tree) {
+    next unless (defined $tree->{$key}{'SPECIES_PRODUCTION_NAME'}); # skip if not a species key
```
ens-ma7, line 894: I am not sure this makes sense since $tree->{$key} should bring back Species production name

ens-ma7:
Another place where i saw the production_name being used differently which might need replaced with your new code @ens-ap5 
https://github.com/Ensembl/ensembl-webcode/pull/809/files#diff-b5af69dd987fd784859294e5e43e8147b36032ad9f25d0415dca71707248320bR809

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -892,124 +907,139 @@ sub _parse {
...
-
-          if ($strain_group && $strain_name !~ /reference/) {
-            $child->{type} = $group_name . ' ' . $tree->{$url}{'STRAIN_TYPE'}. 's';
-          }
-          elsif($strain_group && $strain_name =~ /reference/) {
-            # Create display name for Reference species
-            my $ref_name = $tree->{$url}{'SPECIES_DISPLAY_NAME'} . ' '. $strain_name;
-            $child->{display_name} = $ref_name;
-          }
+      # Populate taxonomy division using e_divisions.json template
```
ens-ma7, line 970: I am not sure what has changed in this part of code since it looks the same just github seems to have shifted everything. Not sure if its the diff thats confused. @ens-ap5 can you confirm if there has been anything changed. It might be that you remove an if statement causing the identation to shift.

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -892,124 +907,139 @@ sub _parse {
...
+            else {
+              $node->{child_nodes} = [];
+              push @{$node->{child_nodes}}, $child;
+              last;
+            }
           }
         }
       }
     }
-    else {
```
ens-ma7, line 1010: I can see this has been removed. so probably an if statement is gone from there. Can you check if thats intended? seems like a valid error to me or maybe not.

modules/EnsEMBL/Web/ConfigPacker.pm
ens-ap5: No - the arrayref will autovivify. What we do need is to remove this, as it duplicates line 1708.

modules/EnsEMBL/Web/ConfigPacker.pm
ens-ap5: Done - but will default $db_name to 'DATABASE_CORE' if undef, so that there's no chance of breaking existing code that might omit the variable.

modules/EnsEMBL/Web/ConfigPacker.pm
ens-ap5: How odd! It looked fine in vim, but there were spaces instead of a line break. I suspect the same formatting exists in eg-web-common - I recall seeing odd indenting in lots of places bc Eugene K didn't set up his .vimrc correctly.

modules/EnsEMBL/Web/ConfigPacker.pm
ens-ap5: Actually we do possibly need to add DB_SPECIES twice, both to the species tree and to the collection tree. It's harmless to non-collection databases and may be needed by code elsewhere.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: We do, but it's not relevant here and it was causing problems with the covid site which now has compara but no ancestral sequence database.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: No, if you look further up the _parse method, you'll see that this tree has other keys - don't confuse $tree here with the tree created by ConfigPacker. See here: https://github.com/Ensembl/ensembl-webcode/blob/feature/collections/modules/EnsEMBL/Web/SpeciesDefs.pm#L802

ens-ap5: Hence you only want the subhashes that contain actual species information. We could use any other relevant key to identify a species subhash, but SPECIES_PRODUCTION_NAME is mandatory and therefore most reliable.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: I think it's the diff that's confusing - I had to break up the loop into two sections in order to add the merging of the generic collection values into the individual species trees at the correct point. Collection databases presumably have additional meta keys, equivalent to DEFAULTS.ini, that apply to all species in the collection.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: See above.

jyothishnt:
@ens-ap5  It is good to test the species selector too. Do you have a sandbox that I can do this?

ens-ap5:
@jyothishnt http://ves-hx2-75.ebi.ac.uk:10104/index.html

ens-ma7 (approved):

ens-ma7:
@ens-ap5 i am happy with the changes made, thanks for these. Is there anything else left for me to review? Dont forget to update the base branch to master for the PR since this is going to master and also do the squash/merge.

PR #812, opened 2021-02-04
Fix GENCODE basic track  - V2

msmanoj:
## Views affected

http://ves-hx2-75.ebi.ac.uk:42229/Homo_sapiens/Location/View?db=core;g=ENSG00000139618;r=13:32315086-32400268

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6132


msmanoj:
Hi @ens-st3 , could you check if it looks okay now, please?

ens-ap5 (approved):
Looks fine to me. You don't actually need the change on line 210, because a) it only applies to the LRG track and b) we don't use this glyphset to draw LRGs in the genome browser any more. It won't break anything, though - maybe I should clean it up in e104 as a separate commit.

PR #813, opened 2021-02-04
Add option to disable tools on mirrors

msmanoj:
## Description

- Added the code back to disable tools on mirrors using the following commits:
-- https://github.com/Ensembl/ensembl-webcode/commit/03bab3dc#diff-67c427902d6bb8985d728ab4d2f1ec075564d3f61960969f4777e33ba1f914a7R108
-- https://github.com/Ensembl/ebi-plugins/commit/b5b7fa2d16a79cee89f1476652727a2a5c22157d

- To disable, we need to create the file `tools_down` in the location: `/nfs/services/nobackup/ensweb/` on the mirrors.

**NOTE:** This needs to be merged along with the following PR:
https://github.com/Ensembl/ebi-plugins/pull/80

## Views affected

http://ves-hx2-75.ebi.ac.uk:42229/info/docs/tools/index.html

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5595

### DIFF
--- a/conf/httpd.conf
+++ b/conf/httpd.conf
@@ -134,6 +134,7 @@ RewriteEngine on
 RedirectMatch 404 ^/sitemaps/.*$
 # Tools redirect
 RewriteCond "%{REQUEST_URI}" ".*Tools/.*"
+RewriteCond /nfs/ensembl_tmp/tools_down !-f
 RewriteRule /(.*) http://www.ensembl.org/$1 [P]
 RewriteCond "%{REQUEST_URI}" ".*/Config/Blast/.*"
 RewriteRule /(.*) http://www.ensembl.org/$1 [P]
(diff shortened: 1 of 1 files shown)


ens-st3 (approved):
The approach of driving this from file is good. @ens-ds23 can you comment on the code Can you create an SOP when this is approved ?

msmanoj:
Hi @ens-ds23 - could you take a look at this PR when you get a chance, please?

ens-ds23:
Sorry about missing this. I get so much noise from github.

This looks ok to me except we need a different path: /nfs/ensembl_tmp doesn't exist on the mirrors. I'd suggest maybe /nfs/services/nobackup/ensweb. Going any deeper than this in the hierarchy (which would be ideal) would need SiteDefs etc which isn't easily available. I think /nfs/services/nobackup/ensweb is good enough for now. It would be nicer not to have a fixed path in the conf file, but I can't think of a good way to do this right now and this would be a good starting place for this anyway.

msmanoj:
Thanks @ens-ds23. I have updated it to look for the `tools_down` in the following dir `/nfs/services/nobackup/ensweb`

PR #815, opened 2021-02-12
Apply orthologues species filters

msmanoj:
## Description

- Added the functionality to filter the orthologues downloads based on the filters applied on the table

## Views affected

http://ves-hx2-75.ebi.ac.uk:42229/Mus_musculus/Gene/Compara_Ortholog?g=ENSMUSG00000017167;r=11:101170523-101190724

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6112

### DIFF
--- a/htdocs/components/20_ComparaOrtholog.js
+++ b/htdocs/components/20_ComparaOrtholog.js
@@ -24,7 +24,6 @@ Ensembl.Panel.ComparaOrtholog = Ensembl.Panel.Content.extend({
   },
 
   updateDownloadButton: function (classNames) {
-    console.log(classNames)
     var downloadButton = $('.export.modal_link')[0];
     var exportButtonUrl = downloadButton.href;
     // Remove the existing filtered_sets param
--- a/htdocs/components/20_ComparaOrtholog.js
+++ b/htdocs/components/20_ComparaOrtholog.js
@@ -20,8 +20,22 @@ Ensembl.Panel.ComparaOrtholog = Ensembl.Panel.Content.extend({
     this.base.apply(this, arguments);
     
     Ensembl.EventManager.register('dataTableFilterUpdated', this, this.updateWithoutOrthologList);
+    Ensembl.EventManager.register('dataTableFilterUpdated', this, this.updateDownloadButton);
   },
 
+  updateDownloadButton: function (classNames) {
+    console.log(classNames)
+    var downloadButton = $('.export.modal_link')[0];
+    var exportButtonUrl = downloadButton.href;
+    // Remove the existing filtered_sets param
+    if(exportButtonUrl.indexOf(';filtered_sets=') > 0){
+      exportButtonUrl = exportButtonUrl.substring(0, exportButtonUrl.indexOf(';filtered_sets='));
+    }
+
+    // Append the new filtered_sets params
+    downloadButton.href = exportButtonUrl + ';filtered_sets=' + classNames.join(',').replace(/\s/g, '');
+
+  },
   updateWithoutOrthologList: function (classNames) {
     var noOrthoList = $('#no_ortho_species li');
     var noOrthoCount = 0;
--- a/modules/EnsEMBL/Web/Component/DataExport.pm
+++ b/modules/EnsEMBL/Web/Component/DataExport.pm
@@ -187,7 +187,12 @@ sub create_form {
       'name'    => 'export_action',
       'value'   => $hub->action,
     },
+    {
+      'name'    => 'filtered_sets',
+      'value'   => $hub->param('filtered_sets') || 'all',
+    }
   ]);
+
   ## Miscellaneous parameters from settings
   my $user_settings = $self->{'viewconfig'}{$hub->param('data_type')}->{_user_settings};
   foreach (@{$settings->{'Hidden'}||[]}) {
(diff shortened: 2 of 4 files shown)


htdocs/components/20_ComparaOrtholog.js
```diff
@@ -20,8 +20,22 @@ Ensembl.Panel.ComparaOrtholog = Ensembl.Panel.Content.extend({
     this.base.apply(this, arguments);
     
     Ensembl.EventManager.register('dataTableFilterUpdated', this, this.updateWithoutOrthologList);
+    Ensembl.EventManager.register('dataTableFilterUpdated', this, this.updateDownloadButton);
   },
 
+  updateDownloadButton: function (classNames) {
+    console.log(classNames)
```
imransl, line 27: Remove `console.log`?

htdocs/components/20_ComparaOrtholog.js
```diff
@@ -20,8 +20,22 @@ Ensembl.Panel.ComparaOrtholog = Ensembl.Panel.Content.extend({
...
+    var downloadButton = $('.export.modal_link')[0];
+    var exportButtonUrl = downloadButton.href;
+    // Remove the existing filtered_sets param
+    if(exportButtonUrl.indexOf(';filtered_sets=') > 0){
+      exportButtonUrl = exportButtonUrl.substring(0, exportButtonUrl.indexOf(';filtered_sets='));
+    }
+
+    // Append the new filtered_sets params
+    downloadButton.href = exportButtonUrl + ';filtered_sets=' + classNames.join(',').replace(/\s/g, '');
```
imransl, line 36: This part is a big confusing. If we are assuming `;filtered_sets=` is going to be there by default, then shouldn't we be checking if it doesn't exist and then add it like this instead:

```
if (exportButtonUrl.indexOf(';filtered_sets=') === -1) {
  exportButtonUrl += ';filtered_sets=';
}

// no need to add/remove filtered_sets a second time
downloadButton.href = exportButtonUrl + classNames.join(',').replace(/\s/g, '');
```

ens-ap5 (changes requested):
Why hasn't the PR template been filled in? I'm not clear on what this commit is doing, since there isn't even a link back to a JIRA ticket.

htdocs/components/20_ComparaOrtholog.js
imransl: After having a look at the code again, seems like you want to remove the everything after `;filtered_sets=` in the string. If that is the case, then we don't need an `if` condition at all. We can use regex to do this one line:

```
// regex passed to .replace will find the sub string ';filtered_sets=' and replace the text after that with empty string
downloadButton.href = exportButtonUrl.replace(/(?<=;filtered_sets=).*/g, '') + classNames.join(',').replace(/\s/g, '');
```

msmanoj:
@ens-ap5  - Sorry, I missed that bit. I have updated the description now.

htdocs/components/20_ComparaOrtholog.js
imransl: After looking at the code again, I realised that you are trying to instead remove what follows after `;filtered_sets=` instead. If this is the case, then we can join lines 32 and 34 without using any `if` conditions to do this:

```
downloadButton.href = exportButtonUrl.substring(0, exportButtonUrl.indexOf(';filtered_sets=')) + ';filtered_sets=' + classNames.join(',').replace(/\s/g, '');
```

msmanoj: I still think it looks cleaner to separate it out into multiple lines. Do we really want to squash it all to a single line?

imransl: Having that as one line is indeed a bit squashed. However, I also feel that if condition is unnecessary as all we need to do is to replace that text whether it has `;filtered_sets=` or not.

A better approach would be to replace at line 29 itself:

```var exportButtonUrl = downloadButton.href.substring(0, exportButtonUrl.indexOf(';filtered_sets='))```

This will make the `if` condition redundant and you won't need to change anything at line 36.

ens-ap5 (approved):

htdocs/components/20_ComparaOrtholog.js
msmanoj: That's done

ens-ap5:
> @ens-ap5 - Sorry, I missed that bit. I have updated the description now.

Thanks! I just want to check that I understood what was going on and that the Perl changes didn't conflict with the fix that I put in before the 103 release

htdocs/components/20_ComparaOrtholog.js
msmanoj: Do you mean `downloadButton.href.substring(0, downloadButton.href.indexOf(';filtered_sets='))` ?

I feel that using `downloadButton.href` twice doesn't look very nice either.

imransl: Yes that's what I meant. Sorry for the typo there. Isn't that still better than having an if condition, which we actually don't need?

Anyway I don't think we need to dwell too much in this, as it is not a major concern. So go ahead with whatever you feel is better :)

msmanoj: In the current example we only use `substring` only if there is a match which I think is slightly more expensive to use. Having an `if` condition avoids the use of `substring` when there is no match so I guess it is better to leave it as it is.

imransl: That is just using one `substring` so very negligible. But let's just leave it like this.

PR #819, opened 2021-03-08
Fix genes in upper pane that are not linkable

msmanoj:
## Description

Fixed the issue with the genes in upper pane that are not linkable. As per the code before this change, the stableId does not getting displayed when the gene description is missing which I think shouldn't be the case.

## Views affected

http://ves-hx2-75.ebi.ac.uk:42228/Camarhynchus_parvulus_GCA_902806625.1/Location/Overview?db=core;r=2:1-152240728;region=2

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6141

### DIFF
--- a/modules/EnsEMBL/Web/ZMenu/Gene.pm
+++ b/modules/EnsEMBL/Web/ZMenu/Gene.pm
@@ -54,13 +54,14 @@ sub _content {
       type  => 'Gene Symbol',
       label => $gene_desc
     });
-
-    $self->add_entry({
-      type  => 'Gene ID',
-      label => $object->stable_id,
-      link  => $hub->url({ type => 'Gene', action => 'Summary' })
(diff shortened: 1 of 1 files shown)


ens-st3:
I can't see gene links in the zmenu I'm afraid

msmanoj:
> I can't see gene links in the zmenu I'm afraid

Hi @ens-st3 - sorry, I missed this one earlier. Could you please check again? Thanks

ens-st3 (approved):
Behaviour is correct. Can you comment on the robustness of the code change @ens-ap5 ?

ens-ap5 (approved):
That looks sensible. It's weird that we would omit the stable ID if there was no description, but I guess when we started developing the www site there were far fewer edge cases!

PR #821, opened 2021-03-17
Bugfix/basenji

ens-ds23:
## Description

Allow multiple instances of certain panels to avoid destructor being called before another instance calls the constructor.
    
## Views affected

Text sequences.

## Possible complications

## Merge conflicts

None

@Ben-Ensembl on slack.


ens-ds23:
Incorrect branch point. Replaced with #822

PR #822, opened 2021-03-17

ens-ds23:
## Description

Allow multiple instances of certain panels to avoid destructor being called before another instance calls the constructor.
    
## Views affected

Text sequences.

## Possible complications

## Merge conflicts

None

@Ben-Ensembl on slack.

### DIFF
--- a/htdocs/components/02_PanelManager.js
+++ b/htdocs/components/02_PanelManager.js
@@ -148,7 +148,7 @@ Ensembl.PanelManager.extend({
    * Creates the panels in the Ensembl object, adds to the panels registry and initializes it
    */
   createPanel: function (id, type, params) {
-    if (this.panels[id]) {
+    if (this.panels[id] && !this.panels[id].multi) {
       this.destroyPanel(id, 'cleanup');
     }
     if (type) {
--- a/htdocs/components/15_TextSequence.js
+++ b/htdocs/components/15_TextSequence.js
@@ -23,6 +23,7 @@ Ensembl.Panel.TextSequence = Ensembl.Panel.Content.extend({
   },
   
   init: function () {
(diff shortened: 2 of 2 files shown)


azangru:
@ens-ds23 Could you add a link to a page that shows the bug?

Also, could you explain, after you've looked into this bug, why it is happening and what the `multi` field is intended to achieve?

I suppose if we do introduce this `multi` field, it should be first set in the parent `Panel` class to false? Part of the object-oriented programming practice that has always scared the hell out of me was how all those fields in inherited objects tend to appear out of nowhere.

ens-ds23:
Here is a page which shows the bug. Note the errors on the console and never-ending spinner. http://www.ensembl.org/Canis_lupus_familiarisbasenji/Gene/Sequence?db=core;g=ENSCAFG00030025836;r=13:58427396-58504058;t=ENSCAFT00030047702;debug=js

multi is needed because the code doesn't sufficiently distinguish class-wide and instance-wide actions. If there is ever a time when there are zero active instances once one has been constructed, the class-wide destructor will be called preventing further instances from being created. On species where the operations are fast this is seen often, almost never for more common species.

multi disables calls to the class-wide destructor. The only time it seems the destructor is called is at page unload anyway. I didn't initialise the field to false in the supertype because I couldn't reliably work out the initialisation order either by reading the code or experimenting and I'm not sure if the presence of a new field will have no negative effects on any panel. (I've checked that it shouldn't for the textsequence panel). I may have another go later today.

azangru:
Can this be closed as superseded by https://github.com/Ensembl/ensembl-webcode/pull/824?

ens-ds23:
Yes @azangru . Closing now.

PR #824, opened 2021-03-24
Initialise randomness for each worker thread.

ens-ds23:
## Description

Initialise randomness in each thread.
    
## Views affected

Text sequences.

## Possible complications

## Merge conflicts

None

@Ben-Ensembl on slack.

### DIFF
--- a/modules/EnsEMBL/Web/Apache/Handlers.pm
+++ b/modules/EnsEMBL/Web/Apache/Handlers.pm
@@ -331,7 +331,8 @@ sub childInitHandler {
   ## This handler gets called by Apache when initialising an Apache child process
   ## @param APR::Pool object
   ## @param Apache2::ServerRec server object
-  ## This handler only adds an entry to the logs
+  ## This handler only adds an entry to the logs and initialises randomness.
+  srand;
   warn sprintf "[%s] Child initialised: %d\n", time_str, $$ if $SiteDefs::ENSEMBL_DEBUG_HANDLER_ERRORS;
 
   return OK;


jgtate (approved):

tuanebi (approved):

azangru:
For future reference — since no Jira — I'll copy relevant parts of our discussions here:

### Immediate cause of the bug on the client
The frontend code creates a `Panel` object for every html element with a class `js_panel`. Every such element will have an id as an html attribute. Every created `Panel` object is then tracked by the `PanelManager` using a dictionary keyed by these ids. When a new `Panel` is registered with `PanelManager`, the `PanelManager` first checks with the dictionary whether a panel with a given id already exists, and [destroys it if it does](https://github.com/Ensembl/ensembl-webcode/blob/release/103/htdocs/components/02_PanelManager.js#L150-L153). If ids are repeated, then the `Panel` objects responsible for other elements get "destroyed", or at least their connection to their corresponding DOM elements gets severed. This results in various bugs.

### Root cause of the bug on the server
Every time the server responds with a sequence, it wraps it in a div with what we hoped to be a randomly generated id ([see source](https://github.com/Ensembl/ensembl-webcode/blob/release/103/modules/EnsEMBL/Web/TextSequence/Output/Web.pm#L176)). These ids turned out to be not so random, because every time an Apache process dies and a child process is spawned, the parent seeds the child's random function with the same seed as it had itself. This results in an identical output of the random function in the child and, consequently, in repeated ids assigned to the sequences returned by the server.

This PR aims to correct this root problem.

azangru (approved):
Can confirm the fix working in my sandbox.

PR #829, opened 2021-04-15
move precache default species to precache plugin Sitedefs

jyothishnt:
## Description

This PR is related to one in ebi-plugins. So Both should be merged together.
Precache runs for a set of species and this should be configurable in Sitedefs.

## Views affected

N/A

## Possible complications

There shouldn't be any complications as this is used only for precache run.

## Merge conflicts

N/A

No JIRA unfortunately.

### DIFF
--- a/modules/EnsEMBL/Web/Query/Generic/Base.pm
+++ b/modules/EnsEMBL/Web/Query/Generic/Base.pm
@@ -91,7 +91,7 @@ sub loop_species {
   my ($self,$args,$subpart) = @_;
 
   my @out;
-  foreach my $sp (qw(Homo_sapiens Mus_musculus Danio_rerio)) {
+  foreach my $sp (@{$SiteDefs::PRECACHE_DEFAULT_SPECIES}) {
     next if ($subpart->{'species'}||$sp) ne $sp;
     my %out = %$args;
     $out{'species'} = $sp;


modules/EnsEMBL/Web/Query/Generic/Base.pm
```diff
@@ -91,7 +91,7 @@ sub loop_species {
   my ($self,$args,$subpart) = @_;
 
   my @out;
-  foreach my $sp (qw(Homo_sapiens Mus_musculus Danio_rerio)) {
+  foreach my $sp (@{$SiteDefs::PRECACHE_DEFAULT_SPECIES}) {
```
tuanebi, line 94: `$SiteDefs::PRECACHE_DEFAULT_SPECIES` is set only in the precachefarm plugins. 
So if the website uses this module there isn't a default value?

jyothishnt: Hi @tuanebi  This is only used by precache (checked code, also confirmed by @ens-ds23) for internal use. So do we need to a default value added to common Sitedefs?

tuanebi: Only precache uses it so adding to common is unnecessary I think

tuanebi (approved):

PR #830, opened 2021-04-15
get alignment from hub

jyothishnt:
## Description

Alignment ID for compara alignments view is stored in sessions. It used to be passed in the url as param "align" before we implemented sessions. You could still pass it via url which gets priority over sessions. Sequence export code only looked at alignment ID from url param. Hub.pm has a method to get alignment id correctly which should be used in this case.

http://ves-hx-78.ebi.ac.uk:9000/index.html

## Views affected

Sequence alignment exports

## Possible complications

Nothing that I am aware of.

## Merge conflicts

N/A

https://www.ebi.ac.uk/panda/jira/browse/ENSWEBREL-1125
https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=492456


ens-ma7:
@jyothishnt can we have a sandbox URL please.

ens-ma7 (approved):

PR #840, opened 2021-07-19
Yearly copyright update

jyothishnt:
Annual copyright update


ens-ap5:
I'm not going to check all 1387 files, but the diff looks good to me!

PR #841, opened 2021-07-27
Changes required to support individual species compara databases

ens-ap5:
## Requirements

## Description

Various changes to bootstrapping code, to configure single-species compara databases

## Views affected

Only relevant to rapid release - should be ignored by other sites

## Possible complications

Unlikely - triggered by the configuration 

`$SiteDefs::SINGLE_SPECIES_COMPARA   = 1;`

which is on a feature branch of ensembl-rapid

## Merge conflicts

Again unlikely - no-one else should be working on this low-level code.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6136


ens-ap5:
104 and master have diverged, so I don't think submitting against postreleasefix/104 is the right thing here

PR #843, opened 2021-07-29
Add support for single-species compara dbs

ens-ap5:
## Requirements

## Description

Add support for single-species compara databases (and also remove some obsolete compara code).

## Views affected

Should only affect the rapid release site, enabling the configuration of the new Homologues view which is found in the ensembl-rapid repository.

## Possible complications

Unlikely, as changes are triggered by the presence of the following configuration in ensembl-rapid:

`$SiteDefs::SINGLE_SPECIES_COMPARA = 1`

## Merge conflicts

Unlikely - the changes are mostly to low-level code.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6136

### DIFF
--- a/modules/EnsEMBL/Web/Object/Gene.pm
+++ b/modules/EnsEMBL/Web/Object/Gene.pm
@@ -752,16 +752,13 @@ sub get_homologies {
   $homology_source      = 'ENSEMBL_HOMOLOGUES' unless defined $homology_source;
   $homology_description = 'ortholog' unless defined $homology_description;
   
-  my $geneid   = $self->stable_id;
   my $database = $self->database($compara_db);
-  my %homologues;
-
   return unless $database;
 
-  my $query_member   = $database->get_GeneMemberAdaptor->fetch_by_stable_id($geneid);
-
+  my $query_member   = $database->get_GeneMemberAdaptor->fetch_by_stable_id($self->stable_id);
   return unless defined $query_member;
   
+  my %homologues;
   my $homology_adaptor = $database->get_HomologyAdaptor;
   my $homologies_array = $homology_adaptor->fetch_all_by_Member($query_member); # It is faster to get all the Homologues and discard undesired entries than to do fetch_all_by_Member_method_link_type
 
--- a/modules/EnsEMBL/Web/Object/Gene.pm
+++ b/modules/EnsEMBL/Web/Object/Gene.pm
@@ -680,38 +680,6 @@ sub get_alternative_locations {
   return \@alt_locs;
 }
 
-sub get_desc_mapping {
-### Returns descriptions for ortholog types.
-### TODO - get this info from compara API
-  my ($self, $match_type) = @_;
-  my %desc_mapping;
-
-  my %orth_mapping = (
-      ortholog_one2one          => '1 to 1 orthologue',
-      apparent_ortholog_one2one => '1 to 1 orthologue (apparent)',
-      ortholog_one2many         => '1 to many orthologue',
-      ortholog_many2many        => 'many to many orthologue',
-      possible_ortholog         => 'possible orthologue',
-  );
-  my %para_mapping = (
-      within_species_paralog    => 'paralogue (within species)',
-      other_paralog             => 'other paralogue (within species)',
-      putative_gene_split       => 'putative gene split',
-      contiguous_gene_split     => 'contiguous gene split',
-  );
-
-  if ($match_type eq 'Orthologue') {
-    %desc_mapping = %orth_mapping;
-  }
-  elsif ($match_type eq 'Paralogue') {
-    %desc_mapping = %para_mapping;
-  }
-  else {
-    %desc_mapping = (%orth_mapping, %para_mapping);
-  }
-  return %desc_mapping;
-}
-
 sub get_homology_matches {
   my ($self, $homology_source, $homology_description, $disallowed_homology, $compara_db) = @_;
   #warn ">>> MATCHING $homology_source, $homology_description BUT NOT $disallowed_homology";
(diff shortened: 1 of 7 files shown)


modules/EnsEMBL/Web/Object/Gene.pm
```diff
@@ -787,16 +752,13 @@ sub get_homologies {
...
-  my %homologues;
-
   return unless $database;
 
-  my $query_member   = $database->get_GeneMemberAdaptor->fetch_by_stable_id($geneid);
-
+  my $query_member   = $database->get_GeneMemberAdaptor->fetch_by_stable_id($self->stable_id);
   return unless defined $query_member;
   
+  my %homologues;
```
ens-ma7, line 761: I am abit confused with this line, you removed the %homologues from line 792 which makes sense but why did you add it back here? It feels like its not being used elsewhere in that function so can be removed unless i missed something.

ens-ap5: You're right! I think I moved it down on the grounds that it's kinda pointless to initialise a variable just before a conditional return, then forgot to check if the variable was actually used. I'll remove it.

modules/EnsEMBL/Web/Query/Availability/Gene.pm
```diff
@@ -44,6 +44,19 @@ sub fixup {
...
+sub _count_homologues {
+  my ($self, $member, $args) = @_;
+
+  my $dbc   = $self->database_dbc($args->{'species'},'compara');
+  my $gmID  = $member->gene_member_id;
+  my $sql   = 'select count(*) from homology_member where gene_member_id = ?';
+  my $sth   = $dbc->prepare($sql);
+  $sth->execute($gmID);
+
+  my @results = @{$sth->fetchall_arrayref||[]};
```
ens-ma7, line 56: Should this line be like this:
`@{$sth->fetchall_arrayref} ||[]` 

feels wrong to dereference an empty array. Also the return line below might produce an error if the array is empty.  can we check if this line is returning the right result?

modules/EnsEMBL/Web/QueryStore/Source/Adaptors.pm
```diff
@@ -37,9 +37,15 @@ sub _database {
   my ($self,$species,$db) = @_;
 
   $db ||= 'core';
-  if($db =~ /compara/) {
-    $species = 'multi';
+
+  ## Rapid site has both single-species and multi-species compara
+  if ($self->{_sd}->SINGLE_SPECIES_COMPARA) {
```
ens-ma7, line 42: How can we test these changes? Can we see in a sandbox? 

Does this affect precache?

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -631,6 +632,7 @@ sub _expand_database_templates {
   ## Autoconfigure databases
   unless (exists $tree->{'databases'} && exists $tree->{'databases'}{'DATABASE_CORE'}) {
     my @db_types = qw(CORE CDNA OTHERFEATURES RNASEQ FUNCGEN VARIATION);
+    push @db_types, 'COMPARA' if $SiteDefs::SINGLE_SPECIES_COMPARA;
```
ens-ma7, line 635: just checking, should the db_types be compara or pan_compara?

modules/EnsEMBL/Web/Query/Availability/Gene.pm
ens-ap5: We should allow for a gene having no homologues, but you're right, that line will fail if the array is empty!

modules/EnsEMBL/Web/QueryStore/Source/Adaptors.pm
ens-ap5: It's tricky to test right now, because I'm waiting for RR compara dbs to be handed over. The new code here will be ignored by all sites apart from RR, and the changes will be tested on RR before that can go live.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: It should be compara - each species has a separate compara database, so it has to be processed with the other species-specific databases. There's also a separate "pan compara" database that contains reference species from www and the NV divisions.

modules/EnsEMBL/Web/QueryStore/Source/Adaptors.pm
ens-ap5: So no, no effect on precache, because RR doesn't have it.

ens-ma7 (approved):

ens-ma7:
I have gone through the code and approved it. I have mostly check the code change looks fine to my best knowledge, we couldn't check on a sandbox. Even though the changes should be for rapid release only, will be good to keep an eye on staging/test (rebuilding packed files for compara) and of course to check the changes are on rapid release (will be good to test on a staging site if there is one for rapid release).

ens-st3:
merge into postreleasefix/104 ?

ens-ap5:
> merge into postreleasefix/104 ?

I thought there might be too many conflicts with my recent changes to SpeciesDefs for an mpush to work smoothly, so it might be best to do it separately

ens-ma7:
> > merge into postreleasefix/104 ?
> 
> I thought there might be too many conflicts with my recent changes to SpeciesDefs for an mpush to work smoothly, so it might be best to do it separately

I have a feeling this might cause problem with mpush with 104 as this will cause postreleasefix/104 branch to be behind.

ens-ma7:
After our discussion in the code review, we suggested to push these commits to master once the release/105 branch is in. And then to use mpush to push the changes to both master and release/105

jyothishnt (approved):
yeah. Since it has been branched of release/104 its better to merge it to the same parent.

modules/EnsEMBL/Web/Object/Gene.pm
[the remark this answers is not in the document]
ens-ap5: I deleted that variable in a commit 2 weeks ago.

PR #845, opened 2021-08-05
add a param to print production name

jyothishnt:
## Description
Add a new argument to print production names which will feed into taxon tree generation script

To be merged with the below PR
https://github.com/EnsemblGenomes/eg-web-common/pull/81

## Views affected
N/A

## Possible complications
Hopefully not

## Merge conflicts
NA

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6325

### DIFF
--- a/utils/dump_ensembl_valid_species.pl
+++ b/utils/dump_ensembl_valid_species.pl
@@ -30,6 +30,7 @@ use LoadPlugins;
 use EnsEMBL::Web::SpeciesDefs;
 
 my $sd = EnsEMBL::Web::SpeciesDefs->new;
-print join "\n", $sd->valid_species, '';
+@sp_list = keys %{$sd->production_name_lookup()};
+print join "\n", @sp_list;
 
 1;


utils/dump_ensembl_valid_species.pl
```diff
@@ -29,7 +30,21 @@ BEGIN
...
+
+GetOptions(
+  "production=s" => \$prod,
+);
+
 my $sd = EnsEMBL::Web::SpeciesDefs->new;
-print join "\n", $sd->valid_species, '';
+my @sp_list;
+
+if ($prod eq 'true') {
```
ens-ma7, line 42: From my understanding here, you have to specify `production=true`. Instead of true, maybe better to have `production=1` and then in the code just check for ($prod). cleaner code and also easier command line

ens-ma7:
I dont understand the purpose of this PR and how its related to the Jira. will discuss in the code review.

utils/dump_ensembl_valid_species.pl
ens-ma7: after discussing with @jyothishnt we decided not to have the extra parameter to dump the species production, in fact the script should be updated to dump the production name by default.

ens-ma7: To also check if the script is used in our release SOP (jira tickets)

jyothishnt: Only used in Release SOP to generate taxon tree.

ens-ma7: Ok thanks for checking, i think we are fine then to go ahead with the suggested changes. Not to have the if statement. And after reviewing Anne PR aswell where she cleanup the species name, its making more sense to remove this one as well.

ens-ma7 (approved):

PR #847, opened 2021-08-16
Fix EG image export issue

msmanoj:
## Description

This PR fixes the image export issue on the EG sites described in the linked ticket.

## Views affected

- Unable to get plants sandbox to work properly.
- I have tried this fix on that staging site.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6339


ens-st3 (approved):
Works on staging, and looks to be a safe code change

PR #849, opened 2021-08-26
Feature/specdef refactor

ens-ap5:
## Description

Final improvements to SpeciesDefs, removing as much duplicate code as possible.

1. Building of ENSEMBL_TAXONOMY_DIVISION factored out into a separate method, as not used by NV (overridden with empty method in eg-web-common)
2. Moved NV-specific deep-copying code into ensembl-webcode so that I can get rid of the giant _parse method in eg-web-common

See also the corresponding PR in eg-web-common.

## Views affected

1. Vertebrate species selector - should remain unchanged after this update. Click on the Select an alignment button on this page:

http://wp-np2-1f.ebi.ac.uk:10105/Mus_musculus/Location/Compara_Alignments/Image?r=4:136366473-136547301

and check that you can select the individual mouse strains in the multiple alignment.

2. Not a view, but NV packed files should be more-or-less the same size after these changes are applied, since the "copy-on-write" code is designed to prevent unnecassary file bloat. This has been checked by comparing freshly-generated files against the equivalent release on /nfs/public/nobackup/ensweb - the file sizes were either identical or within a few bytes, and any differences are likely to be down to using post-104 code with 104 databases.

## Possible complications

Shouldn't have any impact beyond the situations described above.

## Merge conflicts

I don't think anyone else is working on this low-level code.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6168


ens-ma7:
I can see loads of commits which doesnt look right. should this PR be against relase/104 or master?

ens-ap5:
I think the branch is badly out-of-date - I'll look into it and revise the PR

ens-ap5:
> I can see loads of commits which doesnt look right. should this PR be against relase/104 or master?

Doh! Should have been against master.

ens-ma7:
It still looks like there are a lot more commits and commits like this is very strange:
`Merge branch 'master' into feature/bcf`
You should probably check if the branch you created is uptodate with master first before creating the PR.

ens-ma7:
@ens-ap5 when you have time can you have a look at your branch you submitted the PR, its probably outdated from master. Best to merge master into your branch and then do the PR.

ens-ap5:
@ens-ma7 Should I create a new pull request? Merging master again just added another commit.

ens-ma7:
Yes thats probably the safest one but make sure your master is uptodate before creating your new branch.

PR #851, opened 2021-09-06
Feature/sd refactor

ens-ap5:
## Description

Final improvements to SpeciesDefs, removing as much duplicate code as possible.

1. Building of ENSEMBL_TAXONOMY_DIVISION factored out into a separate method, as not used by NV (overridden with empty method in eg-web-common)
2. Moved NV-specific deep-copying code into ensembl-webcode so that I can get rid of the giant _parse method in eg-web-common

See also the corresponding PR in eg-web-common.

## Views affected

1. Vertebrate species selector - should remain unchanged after this update. Click on the Select an alignment button on this page:

http://wp-np2-1f.ebi.ac.uk:10105/Mus_musculus/Location/Compara_Alignments/Image?r=4:136366473-136547301

and check that you can select the individual mouse strains in the multiple alignment.

2. Not a view, but NV packed files should be more-or-less the same size after these changes are applied, since the "copy-on-write" code is designed to prevent unnecassary file bloat. This has been checked by comparing freshly-generated files against the equivalent release on /nfs/public/nobackup/ensweb - the file sizes were either identical or within a few bytes, and any differences are likely to be down to using post-104 code with 104 databases.

## Possible complications

Shouldn't have any impact beyond the situations described above.

## Merge conflicts

I don't think anyone else is working on this low-level code.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6168

### DIFF
--- a/modules/EnsEMBL/Web/SpeciesDefs.pm
+++ b/modules/EnsEMBL/Web/SpeciesDefs.pm
@@ -540,6 +540,8 @@ sub _read_species_list_file {
 
 sub _get_cow_defaults {
 ## Copy-on-write hash (only used by NV)
+## Note: use method instead of an 'our' variable, as the latter can be a pain
+## when shared across plugins
   return {};
 }
 
@@ -1572,11 +1574,17 @@ sub production_name_mapping {
 ### As the name said, the function maps the production name with the species URL, 
 ### @param production_name - species production name
 ### Return string = the corresponding species.url name which is the name web uses for URL and other code
+### Fall back to production name if not found - mostly for pan-compara
   my ($self, $production_name) = @_;
+  my $mapping_name = $production_name;
   
   foreach ($self->valid_species) {
-    return $self->get_config($_, 'SPECIES_URL') if($self->get_config($_, 'SPECIES_PRODUCTION_NAME') eq lc($production_name));
+    if ($self->get_config($_, 'SPECIES_PRODUCTION_NAME') eq lc($production_name)) {
+    $mapping_name = $self->get_config($_, 'SPECIES_URL');
+    last;
   }
+
+  return $mapping_name;
 }
 
 sub assembly_lookup {
--- a/modules/EnsEMBL/Web/SpeciesDefs.pm
+++ b/modules/EnsEMBL/Web/SpeciesDefs.pm
@@ -538,10 +538,19 @@ sub _read_species_list_file {
   return $spp_list;
 }
 
+sub _get_cow_defaults {
+## Copy-on-write hash (only used by NV)
+  return {};
+}
+
 sub _read_in_ini_file {
   my ($self, $filename, $defaults) = @_;
   my $inifile = undef;
   my $tree    = {};
+  
+  ## Avoid deep-copying in NV divisions, to reduce size of packed files
+  ## See https://github.com/EnsemblGenomes/eg-web-common/commit/f702ab75235e66d7e9a979864b858fe0f88485f7
+  my $cow_from_defaults = $self->get_cow_defaults;
   
   foreach my $confdir (@SiteDefs::ENSEMBL_CONF_DIRS) {
     if (-e "$confdir/ini-files/$filename.ini") {
@@ -555,6 +564,7 @@ sub _read_in_ini_file {
       open FH, $inifile or die "Problem with $inifile: $!";
       
       my $current_section = undef;
+      my $defaults_used   = 0;
       my $line_number     = 0;
       
       while (<FH>) {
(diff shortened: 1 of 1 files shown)


modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -538,11 +538,22 @@ sub _read_species_list_file {
...
+}
+
 sub _read_in_ini_file {
   my ($self, $filename, $defaults) = @_;
   my $inifile = undef;
   my $tree    = {};
   
+  ## Avoid deep-copying in NV divisions, to reduce size of packed files
+  ## See https://github.com/EnsemblGenomes/eg-web-common/commit/f702ab75235e66d7e9a979864b858fe0f88485f7
+  my $cow_from_defaults = $self->get_cow_defaults;
```
ens-ma7, line 555: should the function not be `_get_cow_defaults` you are missing an `_` at the beginning

ens-ap5: Thanks for spotting that!

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -1541,11 +1574,18 @@ sub production_name_mapping {
 ### As the name said, the function maps the production name with the species URL, 
 ### @param production_name - species production name
 ### Return string = the corresponding species.url name which is the name web uses for URL and other code
+### Fall back to production name if not found - mostly for pan-compara
   my ($self, $production_name) = @_;
+  my $mapping_name = $production_name;
   
   foreach ($self->valid_species) {
-    return $self->get_config($_, 'SPECIES_URL') if($self->get_config($_, 'SPECIES_PRODUCTION_NAME') eq lc($production_name));
+    if ($self->get_config($_, 'SPECIES_PRODUCTION_NAME') eq lc($production_name)) {
```
ens-ma7, line 1582: sorry one more where we do not need the `lc`

ens-ap5: Well spotted!

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -1541,11 +1574,18 @@ sub production_name_mapping {
 ### As the name said, the function maps the production name with the species URL, 
 ### @param production_name - species production name
 ### Return string = the corresponding species.url name which is the name web uses for URL and other code
+### Fall back to production name if not found - mostly for pan-compara
   my ($self, $production_name) = @_;
+  my $mapping_name = $production_name;
```
ens-ma7, line 1579: is that correct this assignment? If i understand it correctly in case the function doesnt find a corresponding production_name it will return what was sent. should it not return empty or say something "production_name mapping not found"

jyothishnt:
Hi Anne
1. As mentioned in the description, I checked for mouse strains in multiple alignments and I dont see them.
2. Strains are missing in pairwise alignments. I thought we now have HAL alignments for mouse strains?
3. Region comparison (Multi) is not showing patches anymore (http://www.ensembl.org/Homo_sapiens/Location/Multi?r=17:6933406-7436169)

Above are my initial findings while testing your sanbox. I am continuing with the code review.

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -1095,6 +1050,84 @@ sub process_ini_files {
...
+  my $children = [];
+  my $other_species_children = [];
+
+  my @other_species = grep { $_->{key} =~ m/other_species/ } @{$tree->{'ENSEMBL_TAXONOMY_DIVISION'}->{child_nodes}};
+  $other_species[0]->{child_nodes} = [] if ($other_species[0] && !$other_species[0]->{child_nodes});
+
+  my $strain_name = $tree->{$url}{'SPECIES_STRAIN'};
+  my $strain_group = $tree->{$url}{'STRAIN_GROUP'};
+  my $group_name   = $tree->{$url}{'SPECIES_COMMON_NAME'};
+  my $species_key = $tree->{$url}{'SPECIES_URL'}; ## Key on actual URL, not production name
```
jyothishnt, line 1068: we may have to change this to use production name. (I can't remember why I used species url) But this is a different issue that I will be working soon

ens-ap5: OK, I'll leave that one for now!

modules/EnsEMBL/Web/SpeciesDefs.pm
jyothishnt: as the comment says, it needs to fall back to production name if there is no corresponding species url is not found. 
There should always be a species url for a production name. If in any case it didnt find one and returns a production name, will NV continue to work?

ens-ap5: See the comment - there's no URL configured locally for pan compara species, but we don't want this method to return undef, otherwise links to species on other sites won't work. The alternative is to put the fallback into everywhere we create pan_compara links - doable, but probably outside the scope of this PR.

ens-ap5: No, there won't be a species URL for pan compara species, because they're on a different website. For example you might have human homologues to bumblebee on the metazoa pan compara page, but the metazoa site doesn't know that the human url is Homo_sapiens.

ens-ap5:
@jyothishnt 

1. The mouse strain EPO can only be found on mouse aligment pages, not human - this is true on the live site as well
2. No, we don't use HAL any more - and strains are not included in the pairwise alignments on the live site either
3. My sandbox must have been out-of-date - it's showing the patch now.

ens-ma7 (approved):

jyothishnt (approved):

PR #852, opened 2021-09-06
Feature/is strain of

ens-ap5:
## Description

Historically we have used SPECIES_URL as a species lookup key for many compara views, but in many cases this was achieved by doing ucfirst on the production name. We also used a parameter IS_STRAIN_OF which needed to be hardcoded into all relevant ini files. The latter is hard to maintain and the former will cause problems with the rapid release site, where URLs are not simply a sentence-case version of the production name.

We thus need to:
1. change a lot of compara code to use SPECIES_PRODUCTION_NAME as the lookup key
2. remove a lot of ucfirst manipulation of this parameter, and instead use the production_name_mapping() method to determine the URL for a species.

## Views affected

Various compara views:

- Location/Compara_Alignments
- Gene/Compara_Alignments
- Gene/Compara_Ortholog
- Gene/Compara_Paralog
- Gene/Strain_Compara_Orthologs
- Gene/Strain_Compara_Paralogs

Sandbox: http://wp-np2-1f.ebi.ac.uk:10001/Mus_musculus/Gene/Compara_Ortholog?g=ENSMUSG00000017167;r=11:101061349-101081550

I've compared these views against the live site and they look the same.

## Possible complications

Since the changes are largely confined to a few compara components, it should not affect anything else. 

## Merge conflicts

Needs merging asap so that other developers can get on with related tickets.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6090


jyothishnt:
Hi Anne, you could change it to a draft PR if this is not ready to review yet

ens-ap5:
Thanks, @jyothishnt - I hadn't done that before, and the link isn't exactly obvious!

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
```diff
@@ -64,8 +64,8 @@ sub content {
   my $availability = $object->availability;
   my $is_ncrna     = ($object->Obj->biotype =~ /RNA/);
   my $species_name = $species_defs->GROUP_DISPLAY_NAME;
-  my $strain_url   = $species_defs->IS_STRAIN_OF ? "Strain_" : "";
-  my $strain_param = $self->is_strain ? ";strain=1" : ""; # initialize variable even if is_strain is falsy to avoid warnings
+  my $strain_url   = $self->is_strain ? "Strain_" : "";
```
jyothishnt, line 67: On Component.pm ([here](https://github.com/Ensembl/ensembl-webcode/pull/852/files#diff-3169505c843fbc8b3e47d63e013dd0d128cce011375187f2ebb84d7ddc15d595R517)) you mentioned we need to use hub->is_strain. Should we use the same here?

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
```diff
@@ -64,8 +64,8 @@ sub content {
...
+  my $strain_param = $self->is_strain ? ";strain=1" : ""; # initialize variable even if is_strain is false, to avoid warnings
```
jyothishnt, line 68: same here.. `$hub->is_strain`?

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
```diff
@@ -74,53 +74,39 @@ sub content {
...
   foreach my $homology_type (@orthologues) {
     foreach my $species (keys %$homology_type) {
-
-      #do not show strain species on main species view
-      if ((!$self->is_strain && $species_defs->get_config($species, 'IS_STRAIN_OF')) || ($self->is_strain && !$species_defs->get_config($species, 'RELATED_TAXON'))) {
-        delete $not_seen{$species};
-        delete $not_seen{lc $species};
-        next;
+      
+      my $prod_name     = $species_defs->get_config($species, 'SPECIES_PRODUCTION_NAME');
```
jyothishnt, line 86: Are we trying to avoid these calls by using production names? If so should we change https://github.com/Ensembl/ensembl-webcode/blob/release/104/modules/EnsEMBL/Web/Object/Gene.pm#L822 to use production name? Or is it going to be affected in many places?

modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
```diff
@@ -38,7 +38,7 @@ sub get_details {
   my $member = $object->get_compara_Member($cdb);
   return (undef, '<strong>Gene is not in the compara database</strong>') unless $member;
 
-  my $strain_tree  = $self->hub->species_defs->get_config($self->hub->species,'RELATED_TAXON') if($self->is_strain || $self->hub->species_defs->IS_STRAIN_OF || $self->hub->param('strain'));
+  my $strain_tree  = $self->hub->species_defs->get_config($self->hub->species,'RELATED_TAXON') if ($self->is_strain || $self->hub->param('strain'));
```
jyothishnt, line 41: `$hub->is_strain` ? Ignore if we are keeping the method in Component.pm

modules/EnsEMBL/Web/Object.pm
```diff
@@ -343,7 +345,7 @@ sub get_slices {
   foreach (@slices) {
     next unless $_;
 
-    my $name = $_->can('display_Slice_name') ? $_->display_Slice_name : $args->{species};
+    my $name = $_->can('display_Slice_name') ? lc($_->display_Slice_name) : $args->{species};
```
jyothishnt, line 348: `lc`?

jyothishnt (commented):
Couldnt find any major issues. I have added my comments. Thanks

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
ens-ap5: Well caught! I'll fix that.

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
ens-ap5: Yep, thanks!

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
ens-ap5: We probably should, but I think it can be left until 106 - I'll make a note.

modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
ens-ap5: Done.

modules/EnsEMBL/Web/Object.pm
ens-ap5: I have no idea where that came from! I remember debugging some stuff to do with compara slices for the covid site, so maybe it crept in then? I'll delete it.

PR #857, opened 2021-10-05
Use transcript ID in sequence download file name

msmanoj:
## Description
Updated the filename used in sequence download to use transcript ID instead of the symbol.

## Views affected

http://ves-hx2-75.ebi.ac.uk:42229/Homo_sapiens/Transcript/Exons?db=core;g=ENSG00000128573;r=12:50992084-50993185;t=ENST00000403559

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6333

### DIFF
--- a/modules/EnsEMBL/Web/Component/DataExport/Protein.pm
+++ b/modules/EnsEMBL/Web/Component/DataExport/Protein.pm
@@ -73,10 +73,10 @@ sub content {
 sub default_file_name {
   my $self = shift;
   my $name = $self->hub->species;
-  my $data_object = $self->hub->param('g') ? $self->hub->core_object('gene') : undef;
+  my $data_object = $self->hub->param('t') ? $self->hub->core_object('transcript') : undef;
   if ($data_object) {
     $name .= '_';
-    my $versioned_stable_id = $data_object->stable_id_version;
+    my $versioned_stable_id = $data_object->translation_object()->stable_id_version || $data_object->translation_object()->stable_id;
     # Replace '.' with '_' to avoid file extention clashes 
     $versioned_stable_id =~ s/\./_/g; 
     $name .= $versioned_stable_id;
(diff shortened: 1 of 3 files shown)


ens-st3 (commented):
I get the gene stable ID in the protein sequence download (the live site shows gene symbol). I think it should use the versioned translation ID

modules/EnsEMBL/Web/Component/DataExport/Protein.pm
```diff
@@ -76,9 +76,10 @@ sub default_file_name {
   my $data_object = $self->hub->param('g') ? $self->hub->core_object('gene') : undef;
   if ($data_object) {
     $name .= '_';
-    my $stable_id = $data_object->stable_id;
-    my ($disp_id) = $data_object->display_xref;
-    $name .= $disp_id || $stable_id;
+    my $versioned_stable_id = $data_object->stable_id_version;
+    # Replace '.' with '_' to avoid file extention clashes 
+    $versioned_stable_id =~ s/\./_/g; 
+    $name .= $versioned_stable_id;
```
ens-ma7, line 82: maybe double check what data_object returns if there is no version stable id and probably it needs to default to just stable_id (same for transcript.pm below)

ens-ma7: i would leave the previous code as it is and add the following code:
```
    my $versioned_stable_id = $data_object->stable_id_version;
    # Replace '.' with '_' to avoid file extention clashes 
    $versioned_stable_id =~ s/\./_/g; 
     $name .= $versioned_stable_id  || $disp_id || $stable_id;

```

msmanoj: It will return `undef` if there is no versioned stabled ID present. I have updated it to use `stable_id` by default.

msmanoj:
> I get the gene stable ID in the protein sequence download (the live site shows gene symbol). I think it should use the versioned translation ID

@ens-st3 - That's done. Please check again. Thanks.

ens-st3:
> > I get the gene stable ID in the protein sequence download (the live site shows gene symbol). I think it should use the versioned translation ID
> 
> @ens-st3 - That's done. Please check again. Thanks.

Works on the sandbox

ens-ma7 (approved):

PR #862, opened 2021-11-09
Trim whitespace when generating CSVs

azangru:
### Problem 1: "strange CSV"

**Step to reproduce**: Open https://www.ensembl.org/Homo_sapiens/Gene/Matches?db=core;g=ENSG00000139618;r=13:32315086-32400268 and click on the "Download all tables as CSV" button (if is is visible on the page)

**Symptom:** the downloaded CSV wasn't displaying properly in Excel

image

**Likely cause:** sometimes values in a column would start with a line break, making Excel unable to display the whole value properly.

image

**Solution:** always trim white space for values added to CSV columns 

**View after the fix:**

image

### Problem 2: "button disappears after click"

**Steps to reproduce:** click on the "Download all tables as CSV" button (if is is visible on the page) and refresh the page; the button will no longer be visible.

**Cause:**
- the button appears only when a certain counter that keeps track of the number of tables in a panel shows that there are more than one table ([see relevant code](https://github.com/Ensembl/ensembl-webcode/blob/release/104/modules/EnsEMBL/Web/Document/Panel.pm#L457-L470))
- however, component's output can get cached, in which case the table-rendering code will not get run, and the counter will not get incremented ([see relevant code](https://github.com/Ensembl/ensembl-webcode/blob/release/104/modules/EnsEMBL/Web/Component.pm#L229-L236))

**Solution:** no idea, but a naive approach would be to avoid caching a component.

**Notice:** this example shows that any panel with multiple tables will have exactly the same problem. I do not know how best to fix it properly.

## Views affected
http://wp-np2-1e.ebi.ac.uk:8410/Homo_sapiens/Gene/Matches?db=core;g=ENSG00000139618;r=13:32315086-32400268

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6396

### DIFF
--- a/modules/EnsEMBL/Web/Component/Gene/SimilarityMatches.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/SimilarityMatches.pm
@@ -26,6 +26,7 @@ use base qw(EnsEMBL::Web::Component::Gene);
 sub _init {
   my $self = shift;
   $self->cacheable(1);
+  $self->mcacheable(0);
   $self->ajaxable(1);
 }
 
--- a/modules/EnsEMBL/Web/Component/Gene/SimilarityMatches.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/SimilarityMatches.pm
@@ -26,7 +26,6 @@ use base qw(EnsEMBL::Web::Component::Gene);
 sub _init {
   my $self = shift;
   $self->cacheable(1);
-  $self->mcacheable(0);
   $self->ajaxable(1);
 }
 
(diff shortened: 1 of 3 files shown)


ens-ap5 (approved):
Change to table looks good. I guess it won't matter that much if we don't cache that component, but a better solution would be preferable. I can have a look if you like.

azangru:
> I can have a look if you like.

@ens-ap5 That would be great!

ens-ap5:
Problem 2 is non-fixable - the button is created as a result of calls to the new_table method in E::W::Component, but cached panels doesn't re-render the content. A decision was made to remove this code as there's no easy fix and the current behaviour is just confusing.

PR #863, opened 2021-12-03
Show regulation Cell Types menu without Regulatory Build

jcmarca:
## Description
Regulation menus were only shown properly if the species having a Regulatory Build.
We have new data for sus_scrofa, but currently there is no Regulatory Build for it.  
This data is only peaks and signal and the Regulation menu didn't allow to select cell types, experiments, etc without a regulatory Build.
This could be also the case of new species/data comming in the near future.
I make some changes that allows the Regulation menu to be shown if there are cell types in the funcgen DB even if there is no Regulatory Build.

The changes were tested in my Sandbox
http://wp-np2-11.ebi.ac.uk:6008/Sus_scrofa/Location/View?r=1:90744429-90875118

## Views affected
Regulation Menu

_Please provide the URL(s) for any JIRA issues related to this PR._

### DIFF
--- a/modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
+++ b/modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
@@ -1183,7 +1183,6 @@ sub add_regulation_builds {
 
   #######  NOW DO BIG MATRIX NODE!
 
-  my $menu_title    = 'Features by Cell/Tissue';
   my $menu = $reg_menu->append_child($self->create_menu_node('regulatory_features', $menu_title,
       {
         menu   => 'matrix',
--- a/modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
+++ b/modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
@@ -1136,24 +1136,50 @@ sub add_regulation_builds {
   my $key_2         = 'Regulatory_Build';
   my $build         = $data->{$key_2};
   my $type          = $data->{$key_2}{'type'};
-  return unless $type;
 
-  ## Main regulation track - replaces 'MultiCell'
-  my $build_menu  = $reg_menu->append_child($self->create_menu_node('regbuild', 'Regulatory Build'));
-  $build_menu->append_child($self->create_track_node("regulatory_build", "Regulatory Build", {
-    glyphset    => 'fg_regulatory_features',
-    sources     => 'undef',
-    strand      => 'r',
-    labels      => 'on',
-    depth       => 0,
-    colourset   => 'fg_regulatory_features',
-    display     => 'normal',
-    description => $self->databases->{'DATABASE_FUNCGEN'}{'tables'}{'regulatory_build'}{'analyses'}{'Regulatory_Build'}{'desc'}{'reg_feats'},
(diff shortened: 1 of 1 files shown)


modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
```diff
@@ -1136,25 +1136,51 @@ sub add_regulation_builds {
...
+  my $menu_title = 'Features by Cell/Tissue';
+  
+  if (!$type){ #if there is no regulatory build menu it creates a cell/tissue menu 
+	my $cell_type_menu  = $reg_menu->append_child($self->create_menu_node('', ''));
+	$cell_type_menu->append_child($self->create_track_node("", $menu_title, {}));
+  }
+
   #######  NOW DO BIG MATRIX NODE!
 
   my $menu_title    = 'Features by Cell/Tissue';
```
msmanoj, line 1186: Since `$menu_title` is already defined above, this line can be safely removed.

modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
```diff
@@ -1136,25 +1136,51 @@ sub add_regulation_builds {
...
+  
+  return unless @cell_lines;
+
+  my $menu_title = 'Features by Cell/Tissue';
+  
+  if (!$type){ #if there is no regulatory build menu it creates a cell/tissue menu 
+	my $cell_type_menu  = $reg_menu->append_child($self->create_menu_node('', ''));
+	$cell_type_menu->append_child($self->create_track_node("", $menu_title, {}));
+  }
+
```
msmanoj, line 1183: Could you please try removing this `if` block and check if it still works?

modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
jcmarca: Yes, this is a bug, I frogot to remove the old line after move it up.

modules/EnsEMBL/Web/ImageConfigExtension/Tracks.pm
jcmarca: Yes, I tried before without the block but it didn't show the "Features by Cell/Tissue: menu. This is why it was added.

msmanoj (approved):

ens-ma7:
sorry to add my comment to this now but i only had time to look at this now. Just a quick question @jcmarca if i understand correctly now we are going to show cell/ tissues for all species which has a funcgen database. Is that correct? And are we certain that we will have cell/tissues data for all these species by just having a funcgen/regulation database for them.

jcmarca:
Not exactly, not for all species having a funcgen Db. Previously all species having a funcgen Db contained a Regulatory Build, Array mappings or both data. From now on, there will be species containing peak and signal data but not a Regulatory Build. So there will be 3 different cases why a species could have have a funcgen Db. In the new case we should be able to show the cell/tissue menu even if it is not Regulatory Build for this specific species. But in the case of species with only Array mapping data they shoud only show the Oligo probes menu and not the Regulation menu and submenus. Please let me know if you need more clarifications on that.

PR #865, opened 2021-12-03
Feature/species label

ens-ap5:
## Description

Remove ability of the species_label method to accept a production name, as doing the necessary lookup is very expensive. Instead, update code to always pass in a URL - if this needs to be processed multiple times in a loop, create a lookup hash first (outside the loop).

## Views affected

- bacteria BLAST input form
- species selectors for compara views (Alignment Image, Region Comparison)
- other compara views, including zmenus

## Possible complications

There could be other components that pass a production name into species_label - need to check NV repos.

## Merge conflicts

None at the moment.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6434

Bacteria Sandbox URL for testing:
http://wp-np2-1d.ebi.ac.uk:11010 (this is Ridwan's sandbox)

### DIFF
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -192,9 +192,11 @@ sub content {
     }
   } 
   
+  my $lookup = $hub->species_defs->production_name_lookup;
   foreach my $species (sort { ($a =~ /^<.*?>(.+)/ ? $1 : $a) cmp ($b =~ /^<.*?>(.+)/ ? $1 : $b) } keys %orthologue_list) {
     next if $skipped{$species};
     next unless $species;
+    my $species_url = $lookup->{$species};
     
     foreach my $stable_id (sort keys %{$orthologue_list{$species}}) {
       my $orthologue = $orthologue_list{$species}{$stable_id};
@@ -316,7 +318,7 @@ sub content {
       });
 
       my $table_details = {
-        'Species'    => join('<br />(', split(/\s*\(/, $species_defs->species_label($species))),
+        'Species'    => join('<br />(', split(/\s*\(/, $species_defs->species_label($species_url))),
         'Type'       => $self->html_format ? glossary_helptip($hub, ucfirst $orthologue_desc, ucfirst "$orthologue_desc orthologues").qq{<p class="top-margin"><a href="$tree_url">View Gene Tree</a></p>} : glossary_helptip($hub, ucfirst $orthologue_desc, ucfirst "$orthologue_desc orthologues") ,
         'identifier' => $self->html_format ? $id_info : $stable_id,
         'Target %id' => qq{<span class="$target_class">}.sprintf('%.2f&nbsp;%%', $target).qq{</span>},
@@ -349,7 +351,7 @@ sub content {
       sprintf(
         '<p>%d orthologues not shown in the table above from the following species. Use the "<strong>Configure this page</strong>" on the left to show them.<ul><li>%s</li></ul></p>',
         $count,
-        join "</li>\n<li>", sort map {$species_defs->species_label($_)." ($skipped{$_})"} keys %skipped
+        join "</li>\n<li>", sort map {$species_defs->species_label($lookup->{_})." ($skipped{$_})"} keys %skipped
       )
     );
   }   
@@ -377,10 +379,11 @@ sub get_no_ortho_species_html {
   my ($self, $species_to_show, $sets_by_species) = @_;
   my $hub = $self->hub;
   my $no_ortho_species_html = '';
+  my $lookup = $hub->species_defs->production_name_lookup;
 
   foreach (sort {lc $a cmp lc $b} keys %$species_to_show) {
     if ($sets_by_species->{$_}) {
-      $no_ortho_species_html .= '<li class="'. join(' ', @{$sets_by_species->{$_}}) .'">'. $hub->species_defs->species_label($_) .'</li>';
+      $no_ortho_species_html .= '<li class="'. join(' ', @{$sets_by_species->{$_}}) .'">'. $hub->species_defs->species_label($lookup->{$_}) .'</li>';
     }
   }
 
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -377,7 +377,7 @@ sub get_no_ortho_species_html {
   my ($self, $species_to_show, $sets_by_species) = @_;
   my $hub = $self->hub;
   my $no_ortho_species_html = '';
-  my $lookup = $hub->species_defs->production_name_lookup;
+  my $lookup = $hub->species_defs->prodname_to_url_lookup;
 
   foreach (sort {lc $a cmp lc $b} keys %$species_to_show) {
     if ($sets_by_species->{$_}) {
@@ -424,7 +424,7 @@ sub get_export_data {
 
     if (keys %ok_species) {
       # It's the lower case species url name which is passed through the data export URL
-      my $lookup = $hub->species_defs->production_name_lookup;
+      my $lookup = $hub->species_defs->prodname_to_url_lookup;
       return [grep {$ok_species{lc($lookup->{$_->get_all_Members->[1]->genome_db->name})}} @$homologies];
     }
     else {
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -192,11 +192,9 @@ sub content {
     }
   } 
   
-  my $lookup = $hub->species_defs->production_name_lookup;
   foreach my $species (sort { ($a =~ /^<.*?>(.+)/ ? $1 : $a) cmp ($b =~ /^<.*?>(.+)/ ? $1 : $b) } keys %orthologue_list) {
     next if $skipped{$species};
     next unless $species;
-    my $species_url = $lookup->{$species};
     
     foreach my $stable_id (sort keys %{$orthologue_list{$species}}) {
       my $orthologue = $orthologue_list{$species}{$stable_id};
@@ -318,7 +316,7 @@ sub content {
       });
 
       my $table_details = {
-        'Species'    => join('<br />(', split(/\s*\(/, $species_defs->species_label($species_url))),
+        'Species'    => join('<br />(', split(/\s*\(/, $species_defs->species_label($spp))),
         'Type'       => $self->html_format ? glossary_helptip($hub, ucfirst $orthologue_desc, ucfirst "$orthologue_desc orthologues").qq{<p class="top-margin"><a href="$tree_url">View Gene Tree</a></p>} : glossary_helptip($hub, ucfirst $orthologue_desc, ucfirst "$orthologue_desc orthologues") ,
         'identifier' => $self->html_format ? $id_info : $stable_id,
         'Target %id' => qq{<span class="$target_class">}.sprintf('%.2f&nbsp;%%', $target).qq{</span>},
@@ -351,7 +349,7 @@ sub content {
       sprintf(
         '<p>%d orthologues not shown in the table above from the following species. Use the "<strong>Configure this page</strong>" on the left to show them.<ul><li>%s</li></ul></p>',
         $count,
-        join "</li>\n<li>", sort map {$species_defs->species_label($lookup->{_})." ($skipped{$_})"} keys %skipped
+        join "</li>\n<li>", sort map {$species_defs->species_label($_)." ($skipped{$_})"} keys %skipped
       )
     );
   }   
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -423,7 +423,8 @@ sub get_export_data {
 
     if (keys %ok_species) {
       # It's the lower case species url name which is passed through the data export URL
-      return [grep {$ok_species{lc($hub->species_defs->production_name_mapping($_->get_all_Members->[1]->genome_db->name))}} @$homologies];
+      my $lookup = $hub->species_defs->production_name_lookup;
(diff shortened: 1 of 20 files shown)


ens-ap5:
Bacteria BLAST interface:

http://test-bacteria.ensembl.org/Multi/Tools/Blast

Examples of other pages that use the above changes (click on species selector button to see affected interface):

Region Comparison: http://wp-np2-1f.ebi.ac.uk:10105/Homo_sapiens/Location/Multi?db=core;r=17:63992802-64038237;r1=GeneScaffold_4007:494475-539910:1;s1=Ochotona_princeps

Alignments Image: http://wp-np2-1f.ebi.ac.uk:10105/Homo_sapiens/Location/Compara_Alignments/Image?align=1993&db=core&r=17%3A63992802-64038237&r1=GeneScaffold_4007%3A494475-539910%3A1&s1=Ochotona_princeps

ens-ma7:
@ens-ap5 do you have a bacteria sandbox to check

ens-ma7:
I am also adding @jyothishnt since he worked on the compara alignment. @jyothishnt if you have time to have a look since you know more than me whats wrong with the code.

modules/EnsEMBL/Web/Component/Shared.pm
```diff
@@ -946,6 +946,7 @@ sub check_for_missing_species {
   my $db_key        = $args->{cdb} =~ /pan_ensembl/ ? 'DATABASE_COMPARA_PAN_ENSEMBL' : 'DATABASE_COMPARA';
   my $align_details = $species_defs->multi_hash->{$db_key}->{'ALIGNMENTS'}->{$align};
   my $species_info  = $hub->get_species_info;
+  my $url_lookup    = $species_defs->production_name_lookup;
```
ens-ma7, line 949: @ens-ap5 just making sure i understand the function production_name_lookup. It returns a hash with the species production name as key and value the species URL. is that correct?

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -1541,11 +1541,10 @@ sub species_label {
   ### This function will return the display name of all known (by Compara) species.
   ### Some species in genetree can be from other EG units, and some can be from external sources
   ### Arguments:
-  ###     key             String: species production name or URL
+  ###     url             String: species  URL
   ###     no_formating    Boolean: omit italics from scientific name  
-  my ($self, $key, $no_formatting) = @_;
+  my ($self, $url, $no_formatting) = @_;
```
ens-ma7, line 1546: are we sure we are passing the species URL to that function everywhere its being called. For example here:
https://github.com/Ensembl/public-plugins/blob/release/104/tools/modules/EnsEMBL/Web/Object/Blast.pm#L121

That looks like we are passing the production name to that function

modules/EnsEMBL/Web/Component/Shared.pm
ens-ap5: Yes

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: Originally it was URL only, then I changed it to accept production name, which is where the problems started. I'm now going through the code checking for places where we still pass a production name, and if it occurs within a loop, switching to use the production_name_lookup outside that loop.

ens-ma7: @ens-ap5 does that mean you will create a PR for public-plugins as well. This looks like the code needs updating:
https://github.com/Ensembl/public-plugins/blob/release/104/tools/modules/EnsEMBL/Web/Object/Blast.pm#L121

In fact a few places where `species_lable` is being used.

ens-ap5: Sure - I'll check the other repos when I'm done with this one

modules/EnsEMBL/Web/Component/Location/Compara_AlignSliceBottom.pm
```diff
@@ -91,7 +92,7 @@ sub content {
       more_slices     => $i != @$slices,
     });
     
-    my ($species_name, $slice_name) = split ':', $species_defs->production_name_mapping($_->{'name'});
+    my ($species_name, $slice_name) = split ':', $_->{'name'};
```
ens-ma7, line 95: This looks like it should be like this:
`$lookup->{$_->{'name'}}`

modules/EnsEMBL/Web/Component/Shared.pm
```diff
@@ -957,10 +958,10 @@ sub check_for_missing_species {
 
   foreach (keys %{$align_details->{'species'}}) {
     next if $_ eq $species;
-    my $sp_url = $species_defs->production_name_mapping($_);
+    my $sp_url = $url_lookup->{$_};
     if ($align_details->{'class'} !~ /pairwise/
         && ($self->param(sprintf 'species_%d_%s', $align, lc) || 'off') eq 'off') {
-      push @skipped, $_ unless ($args->{ignore} && $args->{ignore} eq 'ancestral_sequences');
+      push @skipped, $sp_url unless ($args->{ignore} && $args->{ignore} eq 'ancestral_sequences');
```
ens-ma7, line 964: dont think you need to change this line, should it still not be `$_` instead of `$sp_url` ? but i vaguely remember that you mention something about this.

modules/EnsEMBL/Web/Object.pm
```diff
@@ -355,7 +355,7 @@ sub get_slices {
       slice             => $_,
       underlying_slices => $underlying_slices && $_->can('get_all_underlying_Slices') ? $_->get_all_underlying_Slices : [ $_ ],
       name              => $name,
-      display_name      => $self->get_slice_display_name($self->hub->species_defs->production_name_mapping($name), $_),
+      display_name      => $self->get_slice_display_name($name, $_),
```
ens-ma7, line 358: are we sure `$name` is the species url name here?

modules/EnsEMBL/Web/SpeciesDefs.pm
```diff
@@ -1574,11 +1573,11 @@ sub species_label {
       ## Pan-compara species - get label from metadata db
       my $info = $self->get_config('MULTI', 'PAN_COMPARA_LOOKUP');
       if ($info) {
-        if ($info->{$key}) {
-          $label = $info->{$key}{'display_name'}
+        if ($info->{$url}) {
```
ens-ma7, line 1576: is that right that we need the species URL as the key here. Before it was the production_name that was the key?

ens-ma7:
@ens-ap5 please dont forget the changes for public-plugins where `species_label` is used, i know they are used for the tools but you might want to check if they are used elsewhere. 
I have also searched the ensembl-genomes repo and it is used a few places, might be code to check them:
https://github.com/search?p=2&q=org%3AEnsemblGenomes+species_label&type=Code

From my understanding of this PR, we are replacing the way `production_name_mapping` function works and also `species_label` works so will be good to search for all the repos in all divisions (vertebrates and nonvertebrates) for these two functions.

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
```diff
@@ -423,7 +424,8 @@ sub get_export_data {
 
     if (keys %ok_species) {
       # It's the lower case species url name which is passed through the data export URL
-      return [grep {$ok_species{lc($hub->species_defs->production_name_mapping($_->get_all_Members->[1]->genome_db->name))}} @$homologies];
+      my $lookup = $hub->species_defs->prodname_to_url_lookup;
+      return [grep {$ok_species{lc($lookup->{$_->get_all_Members->[1]->genome_db->name})}} @$homologies];
```
jyothishnt, line 428: Is `$ok_species` above always going to be lowercased?

modules/EnsEMBL/Web/Component/Gene/FamilyProteins.pm
```diff
@@ -185,19 +185,20 @@ sub content_ensembl {
       { key => 'peptides', title => 'Proteins', width => '80%', align => 'left' },
     );
 
+    my $lookup = $prodname_to_url_lookup;
```
jyothishnt, line 188: `$prodname_to_url_lookup` is not initialised. 
Use `$hub->species_defs->production_name_lookup;` instead?

modules/EnsEMBL/Web/Component/Shared.pm
jyothishnt: We need production name here @ens-ma7 because further down on line 978 where we use this `@skipped` to display a warning we call it like this `$species_defs->species_label($_)` So either we keep production name here or we get rid of the `species_label()` call.

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Compara_Alignments.pm
```diff
@@ -57,18 +58,18 @@ sub json_fetch_species {
     $t->{children}   = [];
     my @children;
 
-    foreach (sort keys %{$row->{'species'}}) {
-      my $url_name  = $hub->species_defs->production_name_mapping($_);
-      my $prod_name = $hub->species_defs->get_config($url_name, 'SPECIES_PRODUCTION_NAME');
+    foreach $_ (sort keys %{$row->{'species'}}) {
+      my ($prod_name, $url_name);
```
jyothishnt, line 62: is it better to keep these declarations before the loop starts?

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Compara_Alignments.pm
```diff
@@ -57,18 +58,18 @@ sub json_fetch_species {
...
+      my ($prod_name, $url_name); 
       my $ancestral = 0;
       if ($_ =~/ancestral_sequences/i) {
+        ## Not a real species, so it's legit to do ucfirst here!
         $url_name = ucfirst($_);
-        $prod_name = $_;
         $ancestral = 1;
       }
-
-      next unless $prod_name;
```
jyothishnt, line 70: are we missing this line?

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Multi.pm
```diff
@@ -121,9 +124,9 @@ sub json_fetch_species {
     }
   }
 
-  if ($shown{$primary_species}) {
-    my ($chr) = split ':', $params->{"r$shown{$primary_species}"};
-    $available_species{$primary_species} = "$species_label - chromosome $chr";
+  if ($shown{$prodname}) {
```
jyothishnt, line 127: Could you please confirm if $shown has prod name as key or is it the url name?
Because the url for Multi is something like this `http://www.ensembl.org/Homo_sapiens/Location/Multi?db=core;r=17:63992802-64038237;r1=17:63470447-63515882:1;r2=GL873537.1:15865824-15911259:-1;s1=Pan_troglodytes;s2=Otolemur_garnettii` where the species s1 and s2 are url names. And this is what is retrieved in line 46 in this file (where %shown is initialised).  May be I have missed something but good to check.

modules/EnsEMBL/Web/Component/Location/Compara_AlignSliceBottom.pm
ens-ap5: That's what I assumed when I did my big update - but when I looked at it on the website, the label was "Ancestral sequence" because it didn't recognise it from the URL. Everything is interconnected with this code, which is why a PR review alone can't guarantee it's correct.

modules/EnsEMBL/Web/Component/Shared.pm
ens-ap5: Jyo is right - the code is correct. We do need the species_label call (to get the final text shown on the web page), and having checked all the compara pages the other day, I made several corrections to fix places where they said "Ancestral species" instead of the species name.

modules/EnsEMBL/Web/SpeciesDefs.pm
ens-ap5: Yes - see comments above.

modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
ens-ap5: I suspect that lc can be removed, but it doesn't seem to be hurting anything for now!

modules/EnsEMBL/Web/Component/Gene/FamilyProteins.pm
ens-ap5: Ah, good catch! Should be "prodnames_to_urls_lookup" (I renamed a couple of methods to make the direction of the mapping clearer).

ens-ap5: I don't think it really matters, since we don't calculate or display gene families any more, but i'll fix it for consistency's sake.

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Compara_Alignments.pm
ens-ap5: It's slightly better practice, yes - I guess i tend to do it because when you're assigning references to a variable you need to be more meticulous about scope. That's less critical here since we're only assigning strings.

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Compara_Alignments.pm
ens-ap5: No, because I don't assign $prod_name until line 72. We always have a production name at this point because that's what the keys of the hash are that we're looping through.

modules/EnsEMBL/Web/JSONServer/SpeciesSelector/Multi.pm
ens-ap5: Ah, good point - we pass the species in as URL names. Will fix this, thanks!

modules/EnsEMBL/Web/Object.pm
ens-ap5: Yes, this is another place where I had to make changes because the web display was wrong.

ens-ma7:
@ens-ap5  what does the function `valid_species` return, is it production_name or species URL name?

ens-ma7:
confirm by @ens-ap5 that valid_species return species URL name which makes the public-plugins code fine (no change needed)

PR #869, opened 2021-12-21
added an example for using rsync with EG FTP site

ens-st3:
## Requirements

## Description

Minor, add an example of using rsync for EG sites

## Views affected

http://wp-np2-1d.ebi.ac.uk:8096/info/data/ftp/rsync.html

## Possible complications

Not tested on a NV sandbox, but there's only one rsync.html so content will be identical with V

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6412

### DIFF
--- a/htdocs/info/data/ftp/rsync.html
+++ b/htdocs/info/data/ftp/rsync.html
@@ -15,17 +15,28 @@ for your downloads.</p>
 in your web browser to locate the files you need, then alter the FTP URL as follows:</p>
 <ul>
 <li>Change the protocol from <kbd>http:</kbd> to <kbd>rsync:</kbd></li>
-<li>Insert <kbd>ensembl</kbd> into the path after the domain name</li>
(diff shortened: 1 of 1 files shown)


htdocs/info/data/ftp/rsync.html
```diff
@@ -15,17 +15,28 @@ <h1>Using rsync to download from the FTP site</h1>
...
 rsync -av rsync://ftp.ensembl.org/ensembl/pub/current_embl/homo_sapiens .
 </pre>
+</li>
+
+<li>The following command will download all the <i>Actinidia chinensis</i> FASTA files from
+    <kbd>http://ftp.ensemblgenomes.org/pub/plants/current/fasta/actinidia_chinensis</kbd> 
+    to the current directory:</p>
+    
+    <pre class="code">
+rsync -av rsync://ftp.ebi.ac.uk/ensemblgenomes/pub/plants/current/fasta/actinidia_chinensis/ .
```
msmanoj, line 37: Hi @ens-st3 - the domain used in the example code block (`ftp.ebi.ac.uk`) here is different from the one on L33 (`ftp.ensemblgenomes.org`). Is this okay?

ens-st3: Old habits, thanks for spotting. I've updated

msmanoj (approved):

PR #870, opened 2022-01-05
Fix orthologues list strain filter

msmanoj:
## Description

`Triticum aestivum` has the strain group value set to `triticum_aestivum` which is causing them to be filtered from getting displayed in the orthologues list.

## Views affected
http://wp-np2-1d.ebi.ac.uk:42228/Triticum_turgidum/Gene/Compara_Ortholog?db=core;g=TRITD2Bv1G183370;r=2B:540943210-541159134;t=TRITD2Bv1G183370.1#list_no_ortho

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6474

### DIFF
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaOrthologs.pm
@@ -95,7 +95,7 @@ sub content {
       }
       else {
         ## Do not show any strain species on main species view
-        if ($strain_group && $strain_group ne lc($species)) {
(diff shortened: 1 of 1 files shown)


ens-st3 (approved):
Functionally this does what I'd expect

ens-ap5 (changes requested):
Please use SPECIES_PRODUCTION_NAME instead of lc($species) - we're trying to get rid of these, not create more!

msmanoj:
> Please use SPECIES_PRODUCTION_NAME instead of lc($species) - we're trying to get rid of these, not create more!

@ens-ap5 - That's done

PR #878, opened 2022-03-06
Fix xss caused by unsanitised url path parameters

azangru:
## Description
Ensembl code does not sanitise url path parameters, which can then me used on the server to inject malignant code into the html of the document.

See [example in production](https://www.ensembl.org/Homo_sapiens/Info/Index/%22%3E%3Cimg%20src%3Dx%20onerror%3Dconfirm(document.domain)%3E)

This PR adds the sanitisation of url path in the relevant Apache handler following one of the suggestions from [this article](https://www.perl.com/pub/2002/02/20/css.html).

## Sandbox url
[link](http://wp-np2-1e.ebi.ac.uk:8410/Homo_sapiens/Info/Index/%22%3E%3Cimg%20src%3Dx%20onerror%3Dconfirm(document.domain)%3E)

## Views affected
All pages that are built using the Legacy.pm template.

## Possible complications
🤷‍♂️ 

Might potentially break any view; but from what I checked (e.g. the Search page), the site is holding up fine.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6493

### DIFF
--- a/modules/EnsEMBL/Web/Apache/SpeciesHandler.pm
+++ b/modules/EnsEMBL/Web/Apache/SpeciesHandler.pm
@@ -23,7 +23,7 @@ use strict;
 use warnings;
 
 use Apache2::Const qw(:common :http :methods);
-use HTML::Entities;
+use HTML::Entities qw(encode_entities);
 
 use EnsEMBL::Web::Exceptions;
 use EnsEMBL::Web::OldLinks qw(get_redirect);
@@ -80,7 +80,7 @@ sub handler {
   my $species         = $r->subprocess_env('ENSEMBL_SPECIES');
   my $path            = $r->subprocess_env('ENSEMBL_PATH');
   my $query           = $r->subprocess_env('ENSEMBL_QUERY');
-  my @path_segments   = map { HTML::Entities::encode($_) } grep $_, split '/', $path;
+  my @path_segments   = map { encode_entities($_) } grep $_, split '/', $path;
 
   # handle redirects
   if (my $redirect = get_redirect_uri($species, \@path_segments, $query)) {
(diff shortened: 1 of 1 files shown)


ens-ap5 (commented):
I'm a bit nervous about merging this in so close to release - it should be safe, but given how crufty the Ensembl codebase is, there could be unforeseen side-effects. Should we wait until we've branched and let it get thoroughly tested on 107 staging?

ens-ap5:
Is there a corresponding JIRA ticket for this work?

azangru:
> Is there a corresponding JIRA ticket for this work?

@ens-ap5 there is, and the link is in the description 🙂

ens-st3:
What are you particularly concerned about @ens-ap5 ?

azangru:
The more I think about my PR, the more uncertain I become about whether the proposed solution is correct.

The specific attack described in the ticket was caused by us using the url pathname as provided by the user when we were generating links to species pages in the header:
- [here](https://github.com/Ensembl/ensembl-webcode/blob/release/105/modules/EnsEMBL/Web/Document/Element/SpeciesBar.pm#L69-L70)
- and [here](https://github.com/Ensembl/ensembl-webcode/blob/release/105/modules/EnsEMBL/Web/Document/Element/SpeciesBar.pm#L111-L119)

I might be able to just sanitise the string only in those two places, which will have very localised consequences. No chance of something else occasionally going wrong, but a decent chance of similar vulnerabilities somewhere else in our code.

jgtate:
I think I would advocate detainting the path components globally, rather than in just the two problem locations that we know about, and for exactly the reason you suggest @azangru : there might well be other places we just don't know about. Using `HTML::Entities::encode` looks sensible to me, though there's a note in that article you quote about `Apache::Util::escape_html` being quicker, which might be important.

jgtate (approved):

azangru:
I checked — we don't have `Apache::Util::escape_html` installed.

PR #881, opened 2022-03-21
FOR 107: add specific regulatory feature lines by type

pblins:
## Requirements

## Description

Add specific regulatory feature lines by type and switch Regulatory features legends to lower case.

E.g.
	•	promoter: 1st line;
	•	enhancer/open chromatin/transcription factor binding: 2nd line;
	•	CTCF: 3rd line.

Discussed with @ens-ap5 

## Views affected

Regulatory build track.
http://wp-np2-25.ebi.ac.uk:6020/Homo_sapiens/Location/View?r=8%3A26291508-26372680

### DIFF
--- a/conf/ini-files/COLOUR.ini
+++ b/conf/ini-files/COLOUR.ini
@@ -574,19 +574,19 @@ _inherit            = fg_regulatory_features
 
 [fg_regulatory_features]
 ctcf                = 40E0D0    CTCF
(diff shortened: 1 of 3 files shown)


ens-ap5:
Hi Paulo
Could we get together to discuss what you're trying to achieve here. The lower-casing of terms is fine, but the changes to the drawing code are not suitable for purpose.
Thanks

PR #882, opened 2022-03-31
Feature/reg tracks

ens-ap5:
## Requirements

## Description

Sorting regulatory features into separate subtracks and lowercasing the labels.

## Views affected

Regulatory features tracks on Region in detail

## Possible complications

Glyphset is shared with Genoverse, which doesn't support subtracks.

## Merge conflicts

Should be fine.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6596

### DIFF
--- a/conf/ini-files/COLOUR.ini
+++ b/conf/ini-files/COLOUR.ini
@@ -574,19 +574,19 @@ _inherit            = fg_regulatory_features
 
 [fg_regulatory_features]
 ctcf                = 40E0D0    CTCF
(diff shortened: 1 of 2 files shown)


ens-ap5:
This branch is OK, but it breaks the regulatory build track on Genoverse, which does not understand subtracks. Therefore it should not be merged until Genoverse has also been fixed. See https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6596

PR #885, opened 2022-04-14
Feature/refactor shared

ens-ap5:
## Description

- Refactoring Component::Shared, which has become something of a dumping ground for miscellaneous methods!
- Refactoring the transcript_table() method, which has turned into spaghetti over the years and has resulted in lots of duplicate code in eg-web-common, as well as making it impossible to extend the code for Rapid Release.

## Views affected
 
* Gene views (upper summary panel)
* Transcript views (upper summary panel)
* Variation and Structural Variation views

## Possible complications

For the most part, methods have simply been moved to more appropriate modules and inheritance paths updated - as long as these pages load without errors, they should function identically.

On the other hand the Gene Summary and Transcript Summary components have been substantially refactored, with all their code moved into a new module, Component::Summary. These views need to be checked across multiple species and divisions and compared with their equivalents on staging to ensure that all data is still be presented.

To check:

- Human
- Mouse
- Mouse strain (any orthologous gene)
- Wheat
- Wheat cultivar (any orthologous gene)

## Merge conflicts

Shouldn't be a problem at this stage.

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6600


ens-ap5:
Region Comparison species selector not working.

PR #886, opened 2022-04-20
Mysql4 gone

ens-st3:
## Description

We are retiring the MySQL 4 branches on Fri 22nd April. This page should be updated on Mon 25th
Please can it be cherry picked onto master too

## Views affected

A static content page only 
http://wp-np2-1d.ebi.ac.uk:8096/info/data/mysql.html 

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6554

### DIFF
--- a/htdocs/info/data/mysql.html
+++ b/htdocs/info/data/mysql.html
@@ -82,7 +82,7 @@ webservice</a>.
 <h2>Summary of servers and ports</h2>
 
 <p>
-<strong>Databases prior to release 48</strong> The MySQL version 4 servers providing access to databases from release 47 and earlier have retired. 
+<strong>Databases prior to release 48</strong>: The MySQL version 4 servers providing access to databases from release 47 and earlier have retired. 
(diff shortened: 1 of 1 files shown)


htdocs/info/data/mysql.html
```diff
@@ -82,14 +82,17 @@ <h2>Biomart</h2>
 <h2>Summary of servers and ports</h2>
 
 <p>
-<strong>Important note:</strong> Because we are serving databases on
-both MySQL4 and MySQL5, not all MySQL instances use the default port;
-please ensure that you specify the correct port when trying to
+<strong>Databases prior to release 48</strong> The MySQL version 4 servers providing access to databases from release 47 and earlier have retired.
```
ens-ma7, line 85: should we not have a column `:` after `Databases prior to release 48`, it feels like its a sub header

ens-ap5 (changes requested):
I agree with Ridwan that it needs a colon in line 85, after the strong tag. Also, the merge should surely be against postreleasefix/106 if it needs to go onto master?

ens-st3:
Change made

PR #897, opened 2022-05-30
Fix the highlighted gene ID in Gene tree

imransl:
## Description

The highlighted gene ID is not displayed correctly in the Highlighted genes info box. This happens when the orthologue gene ID does not have a display name. The reason behind this was that the code checked if the display name of the orthologue exists and when it fails would display the stable ID of the orthologue instead of the current gene.

The changes in this PR will show the stable ID of the current gene when this check fails. It would be ideal to show the display name instead, but I couldn't figure out how to do it as everything I tried ended up breaking the code.

**Sandbox link:** http://wp-np2-1e.ebi.ac.uk:8330/Homo_sapiens/Gene/Compara_Tree?anc=46819155;db=core;g=ENSG00000133392;g1=ENSEEUG00000000659;r=16:15703135-15857028

**For comparison the staging link is:** http://staging.ensembl.org/Homo_sapiens/Gene/Compara_Tree?anc=46819155;db=core;g=ENSG00000133392;g1=ENSEEUG00000000659;r=16:15703135-15857028

## Views affected

Gene tree -> Highlighted gene info box

[ENSWEB-6591](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6591)

### DIFF
--- a/modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
+++ b/modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
@@ -181,23 +181,18 @@ sub content {
 
   # store g1 param in a different param as $highlight_gene can be undef if highlighting is disabled
   my $gene_to_highlight = $hub->param('g1');
-  my $current_gene = $hub->param('g'); # stable ID of current gene
+  my $current_gene_display_name = $gene->Obj->display_xref->display_id;
   my $highlight_gene_display_label;  
      
   my $lookup = $hub->species_defs->prodnames_to_urls_lookup; 
   foreach my $this_leaf (@$leaves) {
     if ($gene_to_highlight && $this_leaf->gene_member->stable_id eq $gene_to_highlight) {
-      $highlight_gene_display_label = $this_leaf->gene_member->display_label;
+      $highlight_gene_display_label = $this_leaf->gene_member->display_label || $current_gene_display_name;
       $highlight_species            = $lookup->{$this_leaf->gene_member->genome_db->name};
       $highlight_species_name       = $this_leaf->gene_member->genome_db->display_name;
       $highlight_genome_db_id       = $this_leaf->gene_member->genome_db_id;
       last;
     }
-  }
-
-  if ($highlight_gene_display_label eq undef) {
-    # $highlight_gene_display_label = $gene->display_name;
-    $highlight_gene_display_label = $current_gene;
   }
 
   # check if highlight ancestor (anc param) is available or not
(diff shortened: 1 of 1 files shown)


modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
```diff
@@ -181,19 +181,25 @@ sub content {
 
   # store g1 param in a different param as $highlight_gene can be undef if highlighting is disabled
   my $gene_to_highlight = $hub->param('g1');
+  my $current_gene = $hub->param('g'); # stable ID of current gene
```
ens-ap5, line 184: Did you try $self->object->gene->display_xref->db_display_name?

You need to check for the display_xref, something like the code here: https://github.com/Ensembl/ensembl-webcode/blob/master/modules/EnsEMBL/Web/Component/Gene/GeneSummary.pm#L51

ens-ap5 (changes requested):
I think we can get the display name of the current gene, since we're on a gene page. I can go through it with you if you're not sure.

modules/EnsEMBL/Web/Component/Gene/ComparaTree.pm
imransl: @ens-ap5 Thanks for the suggestion. using `display_xref->display_id` worked.

ens-ap5 (changes requested):
Judging by other parts of the code, you need to check if there is a display_xref object before calling a method on it. This code will work in human but may throw an error in less well-annotated species.

imransl:
> Judging by other parts of the code, you need to check if there is a display_xref object before calling a method on it. This code will work in human but may throw an error in less well-annotated species.

@ens-ap5 I'm now checking if `display_xref` exists and assign the value of `g` param if it doesn't. Is there a better way to do this?

ens-st3 (commented):
This is a slightly different bug, but very much related. When the focus gene doesn't have a display name, then the info box shows the display name of the gene from the target species, eg http://wp-np2-1e.ebi.ac.uk:8330/Salvator_merianae/Gene/Compara_Tree?anc=46819267;db=core;g=ENSSMRG00000001859;g1=ENSBBBG00000015773;r=QVOM01000133.1:49351-65331
The display is correct when neither have a display name, eg http://wp-np2-1e.ebi.ac.uk:8330/Salvator_merianae/Gene/Compara_Tree?anc=46819384;db=core;g=ENSSMRG00000001859;g1=ENSUAMG00000025427;r=QVOM01000133.1:49351-65331

ens-ap5 (approved):
Looks good!

PR #902, opened 2022-06-30
Change the inheritance of the Transcript component back to Component::Shared

azangru:
Reverts a change made in https://github.com/Ensembl/ensembl-webcode/pull/894 to fix the following bug:

image

**Cause:** Transcript's child `EnsEMBL::Web::Component::Transcript::SimilarityMatches`, uses the `get_matches` method that is declared on `Component::Shared`.

### DIFF
--- a/modules/EnsEMBL/Web/Component/Transcript.pm
+++ b/modules/EnsEMBL/Web/Component/Transcript.pm
@@ -21,7 +21,7 @@ package EnsEMBL::Web::Component::Transcript;
 
 use strict;
 
-use parent qw(EnsEMBL::Web::Component);
+use parent qw(EnsEMBL::Web::Component::Shared);
 
 ## No sub stable_id   <- uses Gene's stable_id
 ## No sub name        <- uses Gene's name


jgtate (approved):

modules/EnsEMBL/Web/Component/Transcript.pm
```diff
@@ -21,7 +21,7 @@ package EnsEMBL::Web::Component::Transcript;
 
 use strict;
 
-use parent qw(EnsEMBL::Web::Component);
```
ens-ma7, line 24: I dont think thats the right fix for this. `Component.pm` is still the parent for this Module not shared.pm

`Shared.pm` is another child of `Component.pm` like `Regulation.pm` or `Variation.pm`. One fix that i can think of is move `get_matches` to `Component.pm` but this will need to look into more as to why this function is missing.

ens-ap5:
I agree with Ridwan. In retrospect, the shared functionality should have 
gone in a helper module, and I’ve done that in my refactor where 
possible, but some of them contain a lot of business logic so it’s 
harder to move that out of the Component inheritance tree.

Anne

azangru:
> Component.pm is still the parent for this Module not shared.pm

If you compare `Component::Transcript` with `Component::Gene`, you will find that `Component::Gene` also [inherits](https://github.com/Ensembl/ensembl-webcode/blob/release/106/modules/EnsEMBL/Web/Component/Gene.pm) from `Component::Shared`.

While I agree that such inheritance doesn't make sense, the purpose of this PR was to fix the broken page discovered after the refactoring in time for the release. I was not planning to do the refactoring in this PR, because I do not sufficiently know the code and have no idea what's going to break. I can close this PR if you'd rather fix this properly.

ens-ap5:
I don’t have a timeline for fixing it properly for the release, as I 
don’t have my laptop back. Maybe we should go with this fix for now and 
look into improving it for 108?

Anne

ens-ma7:
> > Component.pm is still the parent for this Module not shared.pm
> 
> If you compare `Component::Transcript` with `Component::Gene`, you will find that `Component::Gene` also [inherits](https://github.com/Ensembl/ensembl-webcode/blob/release/106/modules/EnsEMBL/Web/Component/Gene.pm) from `Component::Shared`.
> 
> While I agree that such inheritance doesn't make sense, the purpose of this PR was to fix the broken page discovered after the refactoring in time for the release. I was not planning to do the refactoring in this PR, because I do not sufficiently know the code and have no idea what's going to break. I am happy to close this PR if you'd rather fix this properly.

`Component::Gene` doesnt have anything else in the module and it makes sense that it can inherits from `Shared.pm` but with `Transcript.pm` its a different case. Based on the fact that it is just for fixing this issue, I am happy to go with the suggested fix but not sure if this might break other places for transcript.  Probably best to have another ticket for a proper fix (or udpate the same ticket )

azangru:
I was only trying to fix the issue that was picked up by Selenium testing (discussed [here](https://genomes-ebi.slack.com/archives/CKK7EEBA9/p1656517539248979)). Since `Component::Transcript` happily inherited from `Component::Shared` before https://github.com/Ensembl/ensembl-webcode/pull/894, I figured, it can keep inheriting from `Component::Shared` for a bit longer.

PR #903, opened 2022-07-11
UI Fix: names of the samples in 'Configure Track Display'

pblins:
## Requirements

## Description

The names of the samples on the left of the table for Configure Track Display are cut-off. 

Screenshot 2022-07-01 at 15 17 34

Thus, a little more space would was added.

Screenshot 2022-07-11 at 10 00 23

## Views affected

Configure this page -> Regulation -> Feature by Cell/Tissue -> Configure Track Display

[ENSREGULATION-1542](https://www.ebi.ac.uk/panda/jira/browse/ENSREGULATION-1542)


jyothishnt (approved):
Looks ok

PR #904, opened 2022-07-19
Bugfix/off by one error

imransl:
## Description

The regulatory build track starts one off to the left. This has been a longstanding bug and also affected matrix-driven regulation track. I've applied the fix @ens-ap5 recommended for this bug.

### Sandbox URL

http://wp-np2-1e.ebi.ac.uk:8310/Homo_sapiens/Location/View?db=core;g=ENSG00000221914;r=1:54961569-54961649

[ENSWEB-6542](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6542)

## Views affected

Region in detail -> Regulatory build track

Also affects matrix driven regulation tracks.

### DIFF
--- a/modules/EnsEMBL/Draw/Style/Feature/MultiBlocks.pm
+++ b/modules/EnsEMBL/Draw/Style/Feature/MultiBlocks.pm
@@ -60,7 +60,7 @@ sub draw_feature {
   my $x                     = $feature->{'start'} - 1;
   $x                        = 0 if $x < 0;
   $params{'x'}              = $x;
-  $params{'width'}          = $position->{'width'};
+  $params{'width'}          = $position->{'width'} - 1;
   $params{'colour'}         = $feature->{'colour'};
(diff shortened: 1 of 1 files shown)


ens-ap5 (approved):
Looks good!

jyothishnt:
So is it the value coming from the API that is wrong? If so, should we ask them to correct it?

ens-st3 (commented):
See comment in JIRA, is fixing that a case of incrementing width by 1 ?

There are merge commits in this PR ?

imransl:
Thanks for spotting that @ens-st3. I've reduced the width by 1 as you have suggested and it seems to have fixed the issue.

imransl:
> So is it the value coming from the API that is wrong? If so, should we ask them to correct it?

As discussed during the code review yesterday, the correct values are returned by the API and the DB.

ens-ap5 (approved):
Thanks for sorting this out!

ens-st3 (approved):

PR #905, opened 2022-08-02
Parse use_vcf_consequences from json file

likhitha-surapaneni:
### Description

This PR aims to parse a flag `use_vcf_consequences` from the json file which allows the user to fetch consequences from the VCF file instead of calculating on the fly.

### Requires
 https://github.com/Ensembl/ensembl-variation/pull/830

### DIFF
--- a/modules/EnsEMBL/Web/Query/GlyphSet/Variation.pm
+++ b/modules/EnsEMBL/Web/Query/GlyphSet/Variation.pm
@@ -233,6 +233,7 @@ sub fetch_features {
       my $vcf_id = $id;
       $vcf_id =~ s/^variation_vcf_//;
       if(my $vc = $vca->fetch_by_id($vcf_id)) {
+	$no_cons = $vc->{use_vcf_consequences} if defined($vc->{use_vcf_consequences});
         @vari_features = @{$vc->get_all_VariationFeatures_by_Slice($slice, $no_cons)};
       }
     } else {


ens-ap5 (changes requested):
I'm not sure that's the right place to put it, as it will override the limiting by slice size on line 197, which could break the track on large slices. Would be better to put it just before line 196, with an appropriate if block.

likhitha-surapaneni:
Hi @ens-ap5 , thank you for the review.

The intention is to override the limiting the slice size on line 197 if and only if `use_vcf_consequences` is defined in the JSON file. This way, when `use_vcf_consequences` is defined, it fetches consequences for all slice ranges and not just < 1e5. If it is not defined, then the original behaviour is restored. Kindly let me know if this makes sense in terms of memory utilised.

ens-ap5:
Hi Likhitha, have you tested this on your sandbox on slices > 1e5? That would be the real test of whether this approach works.

likhitha-surapaneni:
Hi @ens-ap5 , I tested this on sandbox with and without the flag and it seems to work as expected for slices > 1e5

ens-ap5 (approved):
OK, let's go with this and see what happens!

ens-ap5:
One final thought - is this flag defined in any other species? Putting it onto release/107 means it will eventually get pushed out to all the current sites, not just rapid release

likhitha-surapaneni:
Hi @ens-ap5 , the flag is not defined in any other species i.e all other species are unaffected by the code changes.

PR #906, opened 2022-08-08
Fix id history converter assembly label

imransl:
## Description
The ID History Converter assembly label was showing the genome of chicken which contains a dot (.). This was because when jQuery was querying the class name, which also happen to contain the dot, it couldn't find the element. Apparently, we need to either escape the dot in class names that we used to query with jQuery or query using an attribute name (source: https://stackoverflow.com/questions/32890749/use-a-dot-in-a-class-name-within-jquery-b-avf-toggle). This PR uses the latter, since escaping the dots in strings could become hacky.

### Sandbox URL
http://wp-np2-1e.ebi.ac.uk:8310/Homo_sapiens/Tools/IDMapper

## Views affected
ID History Converter

[ENSWEB-6670](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6670)

### DIFF
--- a/htdocs/components/00_jquery_selecttotoggle.js
+++ b/htdocs/components/00_jquery_selecttotoggle.js
@@ -32,7 +32,10 @@
       // go through all the selectors in the toggleMap and hide them except the one that corresponds to current element's value
       for (var val in toggleMap) {
         if (val !== currValue) {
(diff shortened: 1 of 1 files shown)


ens-ap5 (commented):
The approach sounds good, but I don't feel qualified to review this code.

ens-ma7 (approved):

PR #930, opened 2022-10-26
Fix regulation matrix when feature types are missing

azangru:
## Description
If the regulation database does not contain entries in the `feature_type` table for all different classes of features (Open Chromatin, Histone and Transcription Factor), the regulation matrix becomes unusable. Here's what it looks like (trying to click a checkbox, but nothing happens):

https://user-images.githubusercontent.com/6834224/198000706-0c18d8b3-778e-486f-88de-c846133bcddf.mp4

The reason is because javascript is trying to access a field on an undefined object:

image

The null check added in this PR seems to be sufficient

## Related JIRA Issue
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6517

Sandbox url:
http://wp-np2-1e.ebi.ac.uk:8410/Sus_scrofa/Location/View?r=14:10195351-10284707

_(when testing please check that the sandbox is using the test pig database that doesn't contain all classes of features)_

image

### DIFF
--- a/htdocs/components/25_ConfigRegMatrixForm.js
+++ b/htdocs/components/25_ConfigRegMatrixForm.js
@@ -1406,6 +1406,11 @@ Ensembl.Panel.ConfigRegMatrixForm = Ensembl.Panel.ConfigMatrixForm.extend({
   displayCheckbox: function(obj) {
 
     var data = obj.data;
+
+    if (!data) {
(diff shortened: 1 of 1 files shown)


ens-ap5 (approved):
Looks straightforward enough - what could possibly go wrong? :)

PR #933, opened 2022-11-10
Fix all regulation tracks enabled by default

imransl:
## Description

There is a long list of regulatory tracks that were turned on by default in the Structural Variation page. A side cause of this issue is that the page eventually breaks due to attempting to load a big amount of data into the page.

The issue was fixed after removing `regulatory_features` tracks which were enabled in the code. This was the longest list of regulatory tracks.

Relevant links:

- Live site where the page breaks: https://www.ensembl.org/Homo_sapiens/StructuralVariation/Context?r=13:99984196-99986821;sv=nsv916030;svf=121912273;vdb=variation
- Sandbox where it now works: http://wp-np2-1e.ebi.ac.uk:8310/Homo_sapiens/StructuralVariation/Context?r=13:99984196-99986821;sv=nsv916030;svf=121912273;vdb=variation

## Views affected

Structural variant (genome context)

## Possible complications

Not sure about this yet

[ENSWEB-6624](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6624)

### DIFF
--- a/modules/EnsEMBL/Web/ImageConfig/structural_variation.pm
+++ b/modules/EnsEMBL/Web/ImageConfig/structural_variation.pm
@@ -69,7 +69,7 @@ sub init_cacheable {
   );
 
   $self->modify_configs(
-    [ 'variation_feature_variation', 'somatic_mutation_all', 'regulatory_features', 'functional_other_regulatory_regions' ],
+    [ 'variation_feature_variation', 'somatic_mutation_all', 'functional_other_regulatory_regions'],
     { display => 'normal' }
   );
 


jyothishnt:
Hi @imransl were you able to find which track was breaking the site?

imransl:
> Hi @imransl were you able to find which track was breaking the site?

I am not sure if it is one track or many. Couldn't find out which track/s were the culprit. Only found out that they belong to the `regulatory-features` group, which we don't want to be turned on by default.

ens-ap5 (commented):
Looks good - I thought it might be an ImageConfig thing! We should also check if the other regulatory regions should be turned on by default, but that's less urgent since there are only a few of them.

ens-ap5:
> Hi @imransl were you able to find which track was breaking the site?

I don't think it was any one track, so much as the sheer quantity of them - the regulatory build has increased in size over the years, and probably just hit a tipping point.

ens-st3 (approved):
This allows the page to load, but I think it breaks the loading of regulatory feature evidence tracks. This could be a problem since they are still available in the configuration menu

PR #934, opened 2022-11-16
Add 3D structure links to PDB and AlphaFold in variation page (e109)

nuno-agostinho:
## Description

Add buttons to view 3D protein structure models for PDB and AlphaFold associated with variant consequences, similar to code from https://github.com/Ensembl/public-plugins/pull/555

AlphaFold button is only shown for missense variants.

## Views affected

Variation tab > Genes and regulation > Details of transcript consequences > Overlapping protein domains

For instance: http://wp-np2-11.ebi.ac.uk:6070/Homo_sapiens/Variation/Mappings?db=core;r=8:109335664-109336664;v=rs334;vdb=variation;vf=164384244#ENST00000335295_164384244_C_tablePanel

## Possible complications

No complications, only adds HTML to variation page.

## Merge conflicts

No conflicts.

[ENSVAR-5231](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5231)

### DIFF
--- a/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
@@ -721,8 +721,8 @@ sub detail_panel {
         my $isAFDB = $domain_src =~ '^AFDB-ENSP' && $consequence && $consequence eq 'missense variant';
         if ( $isPDB || $isAFDB ) {
           my $button_url = $hub->url({
-            type    => 'Tools',
-            action  => $isPDB ? 'VEP/PDB' : 'VEP/AFDB',
+            type    => $isPDB ? 'Variation' : 'Tools',
+            action  => $isPDB ? 'PDB' : 'VEP/AFDB',
             var     => $object->name,
             pos     => $tv->translation_start,
             cons    => $consequence,
--- a/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
@@ -750,6 +750,7 @@ sub detail_panel {
             ( $value_url ) = $value =~ /(.+)\./;
           } elsif ($key =~ '^AFDB-ENSP') {
             $key = "ALPHAFOLD";
+            ( $value ) = $value =~ /(.+)\./; # remove chain from AlphaFold ID
             ( $value_url ) = $value =~ /-(.+)-/;
           } elsif ($key eq 'GENE3D') {
             $value_url = "G3DSA:$value" unless $value =~ /^G3DSA:/;
--- a/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
+++ b/modules/EnsEMBL/Web/Component/Variation/Mappings.pm
@@ -711,7 +711,32 @@ sub detail_panel {
         }
       }
       foreach my $domain_src (sort(keys(%domains))) {
-        $data{domains} .= "<b>$domain_src</b><div class=\"column-right\"><ul class=\"compact\">";
+
+        # add button to PDB and AlphaFold models
+        my $structure_button = "";
+        my $consequence = $tv->display_consequence;
+        $consequence =~ s/_/ /g;
+
+        my $isPDB = $domain_src =~ '^PDB-ENSP';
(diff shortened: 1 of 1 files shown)


azangru:
Mmm, the page in the sandbox where the link appears is a variation page (url says `/Homo_sapiens/Variation/Mappings`); but the page with the AlphaFold widget that it links to is a VEP page (url `/Homo_sapiens/Tools/VEP/AFDB`):

Screenshot 2022-11-16 at 13 51 59

Does this look right to you @ens-st3? Not that it's terribly important; just confusing to me. It's kinda like having an AlphaFold page for a variant, but implemented through variant's coordinates (like an AlphaFold page for a VEP result).

nuno-agostinho:
Hey @azangru,

In case this may help clear the URL thing, the AlphaFold widget was made to complement VEP results linked by a button like the ones in the screenshot, so that may be why the AlphaFold widget was put in `/Tools/VEP`. However, as we think these buttons could also be useful in the variation page, we put them here as well.

Regarding the PDB viewer button, I made it point to `/Tools/VEP` for consistency, but I can make it point to `/Variation/PDB` instead.

Maybe in the future, it would be nice to include the AlphaFold model in the variation page, alongside the PDB one... That would make it easier for users to find the AlphaFold models as well when browsing specific variants. 🤔

Best,
Nuno

ens-st3:
I'd advocate changing the link for PDB to `/Variation/PDB`. Using that 'proper' place for it is better than having consistency with Alphafold, which might later move if we choose to add it for Variation pages

nuno-agostinho:
All right, I agree with changing to `/Variation/PDB`. I changed this also in the sandbox in case you want to test. Thanks, @ens-st3!

jgtate (approved):

azangru:
@ens-st3 @nuno-agostinho @jgtate One other thing to notice is that the id for an Alphafold mapping, as seen on the screenshot, is a somewhat fake id that Ensembl assigns to the Alphafold entries. The real alphafold id will not have a `.A` suffix in the end, as you can see on the [Alphafold page](https://alphafold.ebi.ac.uk/entry/P68871) (the suffix is something that Ensembl is adding in its pipelines):

image

Do we want to snip it off before rendering it in html? Do we know if @sgiorgetti is planning to update the pipeline to avoid adding the `.A` suffix?

sgiorgetti:
@azangru - it is not a fake id ... or at least it was not, when Mag agreed with AF to concatenate the chain id (".A") to the file id ("AF-XXX-FY" where "XXX" is the Uniprot accession and "Y" is the fragment identifier/no).
I was not aware of the issue - or it slipped off my RADAR.
Happy to have it removed, if also Variation are OK. @nuno-agostinho - what about the VEP?

nuno-agostinho:
Hey @sgiorgetti, it is okay for VEP. We just have to make sure that AlphaFold parsing in https://github.com/Ensembl/public-plugins/blob/main/tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm#L1272-L1274 is updated accordingly to avoid issues.

azangru:
So here is what we are showing in the VEP table in e108 (snipping away the dot-chain from the id):

image

Compare this with what's being shown It's probably reasonable to keep it consistent with the id shown in the list of mappings:

image

It's probably reasonable to attempt to achieve consistency.

ens-st3:
I checked on the [protein image](http://www.ensembl.org/Homo_sapiens/Transcript/ProteinSummary?db=core;g=ENSG00000139597;r=13:1-50000000;t=ENST00000380130) and we don't show the identifier at all there.

Could you strip of the '.A' etc for this PR @nuno-agostinho ?

ens-st3:
@nuno-agostinho an alternative solution others have discussed is to *add* the chain IDs to VEP. Please hold off on making changes until we can confirm that

ens-st3:
Hi @nuno-agostinho after discussing this further, please can you remove the chain identifier to brig consistency with VEP that way.

nuno-agostinho:
Hey @ens-st3, I removed the chain identifier: http://wp-np2-11.ebi.ac.uk:6070/Homo_sapiens/Variation/Mappings?db=core;r=8:109335664-109336664;v=rs334;vdb=variation;vf=164384244#ENST00000335295_164384244_C_tablePanel

ens-st3 (approved):
Thanks. Ready to merge I think

PR #938, opened 2022-11-17
Fix /ftp proxy to use http

tuanebi:
www.ensembl.org/ftp is not working due to a recent change of http to https. 
Our Apache can't proxy requests to https, as it requires another module installed like mod_ssl.

Need to revert the latest commit.


veidenberg (approved):
That's one of the places where changing to https was not a good idea. Thanks for finding it.

PR #939, opened 2022-11-21
Draw funcgen peak tracks using data stored in bigbed files

azangru:
## Description
Regulation is removing data from the Peak table for some species (human and mouse for e109), and planning to remove the whole table in the future. Instead, they are storing this data in bigbed files; and have introduced an `epigenome_track` table that stores a reference to this file. The purpose of this change, as summarised by Garth, is:

> - Easy to import external bed files, without having to put everything in the peak table
> - To find out whether users need our API and Biomart for peak-based analysis, or whether files are sufficient (in our experience, files are better).
> - To reduce bloat (copying data from release to release even when it's not updated).
> - To harmonise our handover to the new and old sites
> - Biomart was struggling with the number of our peaks 

We are thus in a situation when, for some species, peak track data has still to be read from the database, whereas for other species, it has to be read from bigbed files. This PR introduces support for this.

Done together with @ens-ap5 

## Brief overview of changes
References to bigbed files are stored in the packed files (see ConfigPacker.pm). Then, in the Slice.pm module, the paths to bigbed files are read back from the packed file, and full paths generated. Notice that for species that read peak data from the database, it continues to be read via the peak_calling adaptor.

The track is then drawn by the `fg_multi_wiggle` glyphset. Notice that it has been modified to inherit from multiple parents, one of which has a capability to read data from bigbed files. In the `fg_multi_wiggle` module, track data is read from the bigbed file, and prepared for drawing.

A new zmenu module, `BigbedPeak` has been created specifically for such tracks. The data that is shown in the zmenu is, again, accessed directly from the bigbed file using the url parameters for getting its path. The additional data, belonging to the track as a whole and thus not included in the bigbed file (such as the feature name, or the zmenu title), is accessed via the peak_calling adaptor.  

## Views affected
Region view, configured to show functional genomics tracks.

**Sandbox:** http://wp-np2-1e.ebi.ac.uk:8411/Homo_sapiens/Location/View?r=17:63992754-64038189;db=core (you'll need to configure the page via the regulation matrix to select appropriate tracks) 

## Related JIRA Issues
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6558


ens-ap5 (approved):
Looks good - I think we've already been through most of it together, so I'll trust that it all works :)

PR #942, opened 2022-12-08
New trackhub registry

veidenberg:
This PR will switch trackhub search to the new Trackhub Registry.
I will update the registry URL in the PR once the new service has moved to production.

Sandbox: http://wp-np2-1d.ebi.ac.uk:1680/info/website/trackhubs/index.html
JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6727


veidenberg:
The genecards URL change comes from another PR that is already merged but not yet in postreleasefix branch.

ens-ap5 (commented):
Are we OK with using a "test" URL on the live site? Also, this will need to be merged before the rc sites are set up, otherwise it won't go onto the release/109 branch - I guess it depends on when the new THR goes live?

ens-st3 (approved):

PR #950, opened 2023-02-07
Fix correct tools form options for a particular species not being displayed 

imransl:
## Description
@jyobhai spotted that the relevant options were not being displayed when species are changed in forms when testing the fix for MANE select being displayed to all species (https://github.com/Ensembl/public-plugins/pull/655). This was due to the escape code breaking the species selection.

The related PR in public-plugins is: https://github.com/Ensembl/public-plugins/pull/661

## Views affected
Tools forms (especially Variant Recorder and VEP)

## Possible complications
Some more species selection issues?

[ENSWEB-6716](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6716)


imransl:
I can't update the description due to some issue in GitHub. So posting the correct sandbox link here: https://github.com/Ensembl/ensembl-webcode/pull/950.

jyothishnt (approved):

PR #952, opened 2023-02-10
Fix escapeDot breaking BLAST form

imransl:
## Description

Fix BLAST form being broken because of a change that was made in #950. Now the `className` is explicitly checked if it is undefined or not before the rest of the function is executed.

Sandbox link: http://wp-np2-1e.ebi.ac.uk:8310/Multi/Tools/Blast?db=core

## Views affected

BLAST

## Possible complications

Not sure

[ENSWEB-6794](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6794)


jyothishnt (approved):

azangru (approved):
what could go possibly go wrong? 😃

PR #956, opened 2023-03-28
Jk ensweb 6724

jonk-ebi:
This PR is to update the supported browser list page. The changes can be viewed here http://wp-np2-1f.ebi.ac.uk:10000/info/website/browsers.html. 

The list was generated by looking browser usage stats here https://caniuse.com/usage-table. Safari includes a bigger drop to account for the penultimate version of macOS being locked to version 15.6. 

More details can be seen in the ticket here https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6724

The ticket mentions that molstar should be taken into consideration. Sadly I could not find any minimum browser requirements so I just added a note about hardware acceleration

### DIFF
--- a/htdocs/info/website/browsers.html
+++ b/htdocs/info/website/browsers.html
@@ -18,7 +18,7 @@
   <li>Firefox 108+</li>
   <li>Safari 15.6+</li>
 </ul>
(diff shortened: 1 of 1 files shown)


jonk-ebi:
Updated spelling mistake and ready to be reviewed again

PR #970, opened 2023-06-19
Fix compara ENSWEB-6819

kamaldodiya:
## Requirements

## Description

Fix Compara issue 

## Views affected

Compara

## Possible complications

_We appreciate this can be difficult but please highlight any views that you think might possibly be adversely affected by this change. For example a change to method in ApacheHandlers.pm has potential for widespread consequences._

## Merge conflicts

_At certain stages of the release there is potential for many people to be working on the same modules. To avoid merge conflicts please confirm you have tested for this before submitting the pull request, eg by updating your feature branch and checking that the only changes being submitted are ones from you._

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6819

### DIFF
--- a/modules/EnsEMBL/Draw/GlyphSet_transcript_new.pm
+++ b/modules/EnsEMBL/Draw/GlyphSet_transcript_new.pm
@@ -254,6 +254,10 @@ sub get_gene_connections {
   return unless $ma;
   
   my $qy_member = $ma->fetch_by_stable_id($gene->stable_id);
+  my $species_prod_name = $gene->adaptor->db->get_MetaContainer->get_production_name();
+  my $genome_db = $compara_db->get_GenomeDBAdaptor->fetch_by_name_assembly($species_prod_name);
+  
+  my $qy_member = $ma->fetch_by_stable_id_GenomeDB($gene->stable_id, $genome_db);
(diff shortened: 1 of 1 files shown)


modules/EnsEMBL/Draw/GlyphSet_transcript_new.pm
```diff
@@ -254,6 +254,10 @@ sub get_gene_connections {
   return unless $ma;
   
   my $qy_member = $ma->fetch_by_stable_id($gene->stable_id);
+  my $species_prod_name = $gene->adaptor->db->get_MetaContainer->get_production_name();
+  my $genome_db = $compara_db->get_GenomeDBAdaptor->fetch_by_name_assembly($species_prod_name);
+  
+  my $qy_member = $ma->fetch_by_stable_id_GenomeDB($gene->stable_id, $genome_db);
```
jgtate, line 260: I don't think you need the `my` on this line. It's already declared on 256, so having it here too will generate a warning

kamaldodiya: Well that kine should have been replaced with new lines. Updated.

PR #974, opened 2023-07-11
Fix human self-alignment synteny

azangru:
## Description
The code to display a synteny alignment view is assuming that the alignment can only be done between two different species, and thus breaks when tasked to show a human self-alignment. Meanwhile, we are displaying a link to human self-alignment synteny page on a Compara help&docs page: https://www.ensembl.org/info/genome/compara/analyses.html

The changes in this PR are:
- handle the case when the "main species" is the same as the "other species" in `GlyphSet/Vsynteny.pm`
- make sure Human always shows up int the dropdown of species available for alignment view

## Views affected
Synteny for Human aligned against Human

http://wp-np2-1e.ebi.ac.uk:8410/Homo_sapiens/Location/Synteny?r=8:26291508-26372680;otherspecies=Homo_sapiens

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6820

### DIFF
--- a/modules/EnsEMBL/Web/Component/Location/SyntenyImage.pm
+++ b/modules/EnsEMBL/Web/Component/Location/SyntenyImage.pm
@@ -151,7 +151,7 @@ sub species_form {
   my @values;
 
   foreach my $next (@sorted) {
-    next if $next->{'name'} eq $hub->species and $next->{'name'} ne 'Homo_sapiens'; # Homo sapiens is the only vertebrate that has self-synteny
+    next if ($next->{'name'} eq $hub->species and $next->{'name'} ne 'Homo_sapiens'); # Homo sapiens has self-synteny, so should be visible in the species dropdown even when already on the Human page
     push @values, { caption => $next->{'display'}, value => $next->{'name'} };
   }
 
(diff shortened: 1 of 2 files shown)


jyothishnt (approved):

modules/EnsEMBL/Web/Component/Location/SyntenyImage.pm
```diff
@@ -151,7 +151,7 @@ sub species_form {
   my @values;
 
   foreach my $next (@sorted) {
-    next if $next->{'name'} eq $hub->species;
+    next if $next->{'name'} eq $hub->species and $next->{'name'} ne 'Homo_sapiens'; # Homo sapiens is the only vertebrate that has self-synteny
```
jgtate, line 154: I'm not sure if Perl would correctly intepret the combination of `and` and `or` in that line. I think it would be good to enclose the logic after the "if" in round brackets, in order to avoid the possibility of Perl misinterpreting the priority/binding of the "and". i.e. make the line
`next if ( $next->{'name'} eq $hub->species and $next->{'name'} ne 'Homo_sapiens' );`

PR #984, opened 2023-11-17
Add name field validation

Mehrnaz-Charkhchi:
## Description

Add validation to the name field of registration for to prevent users from using non-ascii characters. 

## Views affected
https://www.ensembl.org/Multi/Account/Register
http://wp-np2-1e:8788/Multi/Account/Register

## Possible complications
Since field type "Name" is added in this process it shouldn't affect other fields.

[ENSWEB-6907](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6907)


htdocs/components/00_jquery_validate.js
```diff
@@ -123,11 +123,13 @@ $.extend($.validator, {
     validClass:    'valid',
     invalidClass:  'invalid',
     requiredClass: 'required',
-    trim: [ 'int', 'nonnegint', 'posint', 'float', 'nonnegfloat', 'posfloat', 'email', 'url' ],
+    trim: [ 'int', 'nonnegint', 'posint', 'float', 'nonnegfloat', 'posfloat', 'email', 'url','name','required' ],
```
azangru, line 126: You probably don't need `required` in this array?

htdocs/components/00_jquery_validate.js
```diff
@@ -123,11 +123,13 @@ $.extend($.validator, {
...
     tests: {
       'int':    new RegExp(/^[\-+]?\d+$/),
       'float':  new RegExp(/^([\-+]?)(?=\d|\.\d)\d*(\.\d*)?([Ee]([\-+]?\d+))?$/),
       password: new RegExp(/^\S{6,32}$/),
+      required: new RegExp(/(.|\s)*\S(.|\s)*/),
```
azangru, line 131: We shouldn't need this field. Tests for required fields were passing even before this PR.

htdocs/components/00_jquery_validate.js
```diff
@@ -123,11 +123,13 @@ $.extend($.validator, {
...
+      name:     new RegExp(/^[\x00-\x7F]+$/),
```
azangru, line 132: Out of interest, did you try running this against some of the names we receive in RT:

- 邮箱未激活
- 张凯强
- 回复

Mehrnaz-Charkhchi: No but I did now. It does return the proper error.

veidenberg (approved):

jyothishnt:
Hi @Mehrnaz-Charkhchi  I searched for `VALIDATION_CLASS` in the entire Ensembl organisation and did not find any place where it is used. Am I missing something?

azangru:
@jyothishnt it is used to add a class name to the form element, which is then targeted by javascript for validation.

See https://github.com/Ensembl/ensembl-webcode/blob/ac2948a8ea595e331c8a06dde69d96d3dea556b6/modules/EnsEMBL/Web/Form/Element/String.pm#L46

jyothishnt:
Aah cool. Thanks I missed that one!

jyothishnt (approved):

htdocs/components/00_jquery_validate.js
jgtate: When I add valid input values in the form on the sandbox the registration button is never enabled. There doesn't appear to be any javascript errors, and javascript seems to be working and it correctly shows an error when I include unicode. Do we know if the button will work in the live site but not in a sandbox, @Mehrnaz-Charkhchi ?

jgtate (approved):
Issue with the register button remaining greyed out is to do with the sandbox site. We'll check it on staging or test before merging to postreleasefix/111 but it looks good otherwise.

jyothishnt:
@Mehrnaz-Charkhchi Any update on this?

PR #989, opened 2023-12-04
Removing GRCh37

olaaustine:
## Requirements

## Description

[ENSVAR-4674](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4674)

## Views affected

https://www.ensembl.org/Homo_sapiens/Location/View?r=19%3A35723836-35723836 
/modules/EnsEMBL/Web/ZMenu/Variation.pm


nakib103 (approved):
Looks good!

PR #990, opened 2023-12-12
Human patches fix

jyothishnt:
## Description

Message from Jose:
The assembly_exception table in the core database used to contain the coordinates of the patches with respect to the corresponding reference chromosome. This was because patches were represented as fake chromosomes in the genome browser. Since release 110, patches are on its own, so the assembly exceptions don’t make sense anymore and have been deleted from the database.
I am not sure about the implications of this in the context of the webcode. Perhaps the patch start should always be 1 and the patch end be the patch length in this new scenario?

Adjusting the patch start and end fixed.

## Views affected

Region comparison and Region in detail

Sandbox for testing http://wp-np2-1e.ebi.ac.uk:9000/

## Possible complications
Dont know

https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6751
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6915
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6887

### DIFF
--- a/modules/EnsEMBL/Draw/GlyphSet/patch_ref_alignment.pm
+++ b/modules/EnsEMBL/Draw/GlyphSet/patch_ref_alignment.pm
@@ -36,12 +36,8 @@ sub _init {
   my $join_colour   = $self->my_colour(lc $self->{'container'}->assembly_exception_type, 'join'); 
 
   if (scalar @$features) {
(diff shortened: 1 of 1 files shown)


jgtate (approved):

jmgonzmart (approved):
These changes look correct to me since there are no assembly exceptions in the core database any more. Tested in the sandbox provided by Jyo, the primary assembly mapping track is displayed correctly.

PR #1017, opened 2024-06-05
Add method to enable Cactus alignment to be hidden

twalsh-ebi:
## Requirements

## Description

In conjunction with [eg-web-plants PR 65](https://github.com/EnsemblGenomes/eg-web-plants/pull/65), this PR (for the `release/112` branch only) makes it possible to suppress the Wheat Cactus alignment on the Ensembl Plants 112 site.

To take effect, it is necessary to repack the `MULTI` packed file.

## Views affected

This change has the effect of making the Wheat Cactus alignment disappear from the Wheat genomic alignment dropdown...

genomic_alignment_dropdown_without_wheat

...as well as from the multiple genomic alignment list:

genomic_alignment_listing_without_wheat

Examples:
- [sandbox](http://wp-np2-25.ebi.ac.uk:5098/Triticum_aestivum/Location/View?r=3D:2585940-2634711)
- [live site](https://plants.ensembl.org/Triticum_aestivum/Location/View?r=3D:2585940-2634711)

## Possible complications

No complications are expected. The `ConfigPacker::_alignment_mlss_is_suppressed` method returns `0` by default, so it should have no effect except when overridden.

## Merge conflicts

None detected.

- [ENSWEB-6910](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6910)

### DIFF
--- a/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -1178,6 +1178,7 @@ sub _build_compara_default_aligns {
     $sth->bind_param(2,$species);
     $sth->execute;
     my ($mlss_id)= $sth->fetchrow_array;
+    die "MLSS $mlss_id cannot be a default Compara alignment, it is suppressed" if $self->_alignment_mlss_is_suppressed($mlss_id);
     push @defaults,[$mlss_id,$species,$method];
     $sth->execute;
     $sth->finish;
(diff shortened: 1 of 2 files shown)


twalsh-ebi:
Hi @jyothishnt.. I'd have no hesitation in acknowledging that this solution is a little ad-hoc, though I reasoned that it might be OK given that it would be specific to `release/112`.

For `release/113` and beyond we are pursuing two alternative approaches to addressing the issues affecting the Wheat Cactus, so it is very unlikely that the Wheat Cactus alignment will need to be hidden in this way in future.

twalsh-ebi:
Hi @jyothishnt.. I have added example links to this PR and [eg-web-plants PR 65](https://github.com/EnsemblGenomes/eg-web-plants/pull/65). In the sandbox example you can see that the Wheat Cactus alignment has... disappeared.

Please let me know whether you have any outstanding concerns relating to this PR; if so, I would be very grateful for the opportunity to address them.

PR #1028, opened 2024-07-29
Increase Pig breed count in Vertebrates genome list

twalsh-ebi:
## Requirements

## Description

This PR increases the Pig breed count in the Vertebrates genome list to match the number of Pig breeds represented in Ensembl 113 (less the Pig reference genome).

## Views affected

This affects the genome list view in the Ensembl Vertebrates homepage.

## Possible complications

This is a straightforward HTML text update, so no complications are expected. 

## Merge conflicts

None detected.

N/A


azangru (approved):

EreboPSilva (approved):
looks good to me

PR #1052, opened 2024-10-24
Update rapid site retirement message

jyothishnt:
Rapid retirement message
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2608

For Rapid site banner, we just need to create a text file with name ensembl_tmp_message on to tmp dir root with below content.

`banner_message <h3 style="color:red;">We will be retiring Ensembl Rapid Release</h3><div>All Rapid Release data, including release 65, has been uploaded into the new Ensembl Beta site. </div><div>New genome data will only be released to the new Ensembl Beta site. We will no longer release data to the Ensembl Rapid Release browser.</div><br><div><b>Go to <a href="http://beta.ensembl.org" target="_blank">beta.ensembl.org</a> now</b></div><br><div>If you feel that this change will have a significant impact on your analyses, please <a href="https://rapid.ensembl.org/Help/Contact">get in touch</a></div>
class banner_msg`

### DIFF
--- a/htdocs/components/99_announcement_banner.css
+++ b/htdocs/components/99_announcement_banner.css
@@ -40,3 +40,10 @@
     margin:0;
     font-weight: normal;
 }
+
+
+.banner_msg {
+  background-color: #f7e4b7;
+  text-align: center;
+  padding: 16px;
+}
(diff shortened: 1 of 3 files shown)


azangru (approved):
Technically, if we needed to override standard classes on a one-off basis, we could have just used inline styles in the html of the banner, couldn't we? I don't know if that would have been better or worse than the proposed solution though; so happy to approve.

jyothishnt:
I wouldn't worry tbh because the styles added here is very simple and probably going to be the same for any banners. So let's leave it and worry when there is a change needed.

jyothishnt:
**PLEASE DO NOT MERGE THIS PR UNTIL WE ARE READY TO RELEASE THE BANNER**

PR #1064, opened 2025-03-24
Migrate precache script to slurm

veidenberg:
- Migrate precache generation script from using LSF to Slurm for job submission
- Update datafiles root path

Related tickets: [ENSWEBREL-1018](https://www.ebi.ac.uk/panda/jira/browse/ENSWEBREL-1018), [ENSWEBREL-1056](https://www.ebi.ac.uk/panda/jira/browse/ENSWEBREL-1056)

### DIFF
--- a/conf/SiteDefs.pm
+++ b/conf/SiteDefs.pm
@@ -43,7 +43,6 @@ our $ENSEMBL_RELEASE_DATE   = 'January 2025'; # As it would appear in the copyri
 ###############################################################################
 ## Default folder locations
 our $ENSEMBL_SERVERROOT   = _get_serverroot(__FILE__);              # Root dir that contains all Ensembl checkouts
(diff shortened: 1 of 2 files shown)


TamaraNaboulsi (approved):
Converting the LSF bsub command to SLURM srun. Time and memory parameters seem adequate. Looks good to me.

EbiArnie (approved):
The whole precache-farm script should be rewritten.
It does not make any sense to use Parallel::Forker to wrap srun commands.
We should get rid of parallel forker and related code and just use sbatch.
But that would be a minor rewrite, so OK for now, with a wish for refactoring later.

veidenberg:
Datafile root path udpate reverted in eb9bb58 (as it's already been set in [ebi-plugins](https://github.com/Ensembl/ebi-plugins/blob/e03ed317e49936340102e4b6e9c43a133c268cf9/common/conf/SiteDefs.pm#L44))

PR #1078, opened 2025-04-30
[115] Hide unneeded ancestral genome sequence

twalsh-ebi:
## Description

PR [ensembl-webcode PR 155](https://github.com/Ensembl/ensembl-webcode/pull/155) implemented functionality to "Remove the unneeded ancestral species from the sequence view".

The intended behaviour was that in ancestral alignment views (e.g. EPO), hiding all the extant genomes descending from a child of an ancestral sequence would result in that ancestral sequence also being hidden.

The most recent release with this behaviour was Ensembl 104: https://may2021.archive.ensembl.org/Homo_sapiens/Location/Compara_Alignments/Image?align=1944&db=core&r=17%3A63992802-64038237

This pull request would restore that behaviour.

## Views affected

This change would affect EPO alignments in Ensembl Vertebrates.

The following example images show the effect of the change when Mouse lemur is hidden in the Primates EPO alignment views.

Example image alignment view before change (with unneeded ancestral sequence displayed):
primates_epo_image_alignment_before

Example image alignment view after change:
primates_epo_image_alignment_after

Example text alignment view before change (with unneeded ancestral sequence displayed):
primates_epo_text_alignment_before

Example text alignment view after change:
primates_epo_text_alignment_after

## Possible complications

This PR would change the return value of method `EnsEMBL::Web::Object::get_alignments` to include the relevant `$align_slice` object. All known invocations of this method have been adapted in the PR, but it's possible that other invocations of the method would need to be adapted in response to the change in this method's behaviour. 

## Merge conflicts

None detected.

N/A

### DIFF
--- a/modules/EnsEMBL/Web/Command/DataExport/Output.pm
+++ b/modules/EnsEMBL/Web/Command/DataExport/Output.pm
@@ -381,7 +381,7 @@ sub write_alignment {
       else {
         $self->object->{'alignments_function'} = 'get_SimpleAlign';
 
-        $alignment = $self->object->get_alignments({
+        ($alignment) = $self->object->get_alignments({
           'slice'     => $data->slice,
           'align'     => $hub->param('align'),
           'species'   => $hub->species,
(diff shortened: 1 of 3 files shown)


twalsh-ebi:
I've marked this as a draft because further testing has uncovered that exported alignment files still contain the unneeded ancestral genome sequence, even if they are hidden from the image/text alignment view.

twalsh-ebi:
Display names for extant genomes have been reverted to use the `species.url`, because although text alignment views may be considered to work better with display names in some cases ...
primates_epo_display_names

... the potential use of very long display names in text alignments may result in the alignment data being "pushed" out of view in other cases:
dmel_pangenome_display_names

PR #1086, opened 2025-06-03
Changes to switch from VEP to Ensembl VEP as per requirements

andrewyatz:
## Description

As per our guidelines the Ensembl Variant Effect Predictor can be referred to either as 

- Ensembl Variant Effect Predictor,
- Ensembl Variant Effect Predictor (VEP)
- Ensembl VEP

No other names are allowed. This follows on from additional work by variation to reduce the number of places we do not conform to guidelines. Additional PRs are being created and will be linked to here.

- Public plugins: public-plugins#880
- EBI plugins: ebi-plugins#192

Specific changes made

- Name usage switched to either Ensembl VEP, Ensembl Variant Effect Predictor (VEP) or just Ensembl Variant Effect Predictor
- FTP table now renders "Ensembl VEP" for column and each download link
- Tools render correctly

## Views affected

- Index home page for www, grch37
- Footer on any site using FatFooter Perl module
- FTP table
- Tools table

## Possible complications

None known.

## Merge conflicts

None found

None

### DIFF
--- a/htdocs/info/website/upload/var.html
+++ b/htdocs/info/website/upload/var.html
@@ -14,7 +14,7 @@
 </ul>
 
 <p>
-The <a href="/info/docs/tools/vep/index.html">Variant Effect Predictor</a>
+The <a href="/info/docs/tools/vep/index.html">Ensembl Variant Effect Predictor (VEP)</a>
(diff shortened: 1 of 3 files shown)


sgiorgetti (approved):
OK

PR #1091, opened 2025-06-23
Match highlighted orthologue by gene and species

twalsh-ebi:
## Requirements

## Description

This PR would:
- pass an `s1` parameter with the gene-tree links in an orthologue table, whose value is the URL of the species containing the given orthologue;
- add an `s1` parameter to gene-tree views, which is expected to contain the species URL of the genome containing the gene specified by `g1`;
- use the species indicated by `s1` to help unambiguously identify the gene specified by `g1` for highlighting in a given gene-tree view; and
- repurpose the `'_leaves'` item of node features in the `genetree` GlyphSet to identify leaves by their concatenated `genome_db_id` and gene stable ID, making it possible for a highlighted orthologue to be identified even if in a collapsed branch of the gene tree.

## Views affected

This is expected to affect gene-tree views in which an orthologue is highlighted, such as those linked from the `ComparaOrthologs` view.

## Possible complications

Complications are not expected, as the `s1` parameter has not (as far as I know) been used in a `ComparaTree` component, and the `'_leaves'` item of node features should be safe to repurpose as they are not (as far as I know) currently used.

## Merge conflicts

None detected.

- TBD

### DIFF
--- a/modules/EnsEMBL/Draw/GlyphSet/genetree.pm
+++ b/modules/EnsEMBL/Draw/GlyphSet/genetree.pm
@@ -208,10 +208,10 @@ sub _init {
     }
 
     if ( $f->{label} ) {
-      if( $other_gene && $f->{_genes}->{$other_gene} ){
+      if( $other_gene && $other_genome_db_id && $f->{_leaves}->{$other_genome_db_id . '|' . $other_gene} ){
         $bold = 1;
         $label_colour = "ff6666";
-      } elsif( $f->{_genes}->{$current_gene} ){
+      } elsif( $f->{_leaves}->{$current_genome_db_id . '|' . $current_gene} ){
         $label_colour     = 'red';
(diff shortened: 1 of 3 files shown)


modules/EnsEMBL/Draw/GlyphSet/genetree.pm
```diff
@@ -746,7 +746,7 @@ sub features {
     $f->{'_height'}             = 12 * log($f->{'_collapsed_count'});
     $f->{'_genome_dbs'}         = \%genome_dbs;
     $f->{'_genes'}              = \%genes;
-    $f->{'_leaves'}             = \%leaves;
```
twalsh-ebi, line 749: For what it's worth, I've checked the webcode repos and failed to find `_leaves` being used as a hash key anywhere other than the `EnsEMBL::Draw::GlyphSet::genetree` module.

PR #1155, opened 2026-04-09
Use Mail::Address methods instead of Email::Address::XS

TamaraNaboulsi:
## Description

A fix to a problem created by #1145 where there was a complication in installing Email::Address::XS. The solution devised aims to use methods from Mail::Address instead, which is a package that comes with Mail::Mailer which is already installed.

## Views affected

The sending helpdesk email form.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -11,7 +11,6 @@ requires 'DB_File';
 requires 'DBD::mysql';
 requires 'DBI';
 requires 'Digest::MD5';
-requires 'Email::Address::XS';
 requires 'File::Spec::Functions';
 requires 'GD::Simple';
 requires 'Hash::Merge';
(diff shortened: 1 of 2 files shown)


veidenberg (approved):
Assuming it's tested (no sandbox URL), it looks good to me 👍

TamaraNaboulsi:
> Assuming it's tested (no sandbox URL), it looks good to me 👍

Thanks Andres. I tested it locally and on one of the staging sites. I don't have a sandbox set up, but if there's one ready somehwere I can test it there.

TamaraNaboulsi:
When testing on a sandbox, email was not being sent. Turns out there was a typo in a regex that checks the url and the presence of dashes '-' was triggering a warning. Fixed in last commit.

PR #1169, opened 2026-06-17
Access pairwise alignment metadata via Multi packed file

twalsh-ebi:
## Description

This PR would change `ConfigPacker::_summarise_compara_db` to pack `SYNTENIES` metadata, and would update `EnsEMBL::Web::Document::HTML::Compara::pairwise_mlss_data`  to use packed `ALIGNMENTS` and `SYNTENIES` metadata instead of the Compara Perl API when generating the pairwise genomic alignments/syntenies list.

## Views affected

This will change how alignment information is fetched for the Vertebrates pairwise genomic alignments page.

It is not expected to change what alignments are presented, though it may change the order in which they are presented.

Example: [sandbox](http://wp-np2-35.ebi.ac.uk:5092/info/genome/compara/analyses.html) vs [live site](https://ensembl.org//info/genome/compara/analyses.html)

## Possible complications

This change would require repacking of all MULTI.db.packed files.

## Merge conflicts

None detected.

- ENSCOMPARASW-9255

### DIFF
--- a/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -1244,7 +1244,7 @@ sub _summarise_compara_db {
       where mlss.method_link_id = ml.method_link_id and
         mlss.species_set_id = ss.species_set_id and 
         ss.genome_db_id = gd.genome_db_id and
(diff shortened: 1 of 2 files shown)


sgiorgetti (approved):
LGTM - thanks

